PR #10, opened 2022-02-17
Invoke Web Assembly-specific dependencies when building for Wasm.

jasonjgw:
The purpose of the  proposed change is to allow MatCAT to be compiled to Web Assembly without modification.

This isn't compiling yet. Changes may be necessary before it can be merged.

On my system, at least, compiling dependencies that include C code fails, as Clang can't find the Standard C library (specifically, a file not found error is returned when opening stdlib.h).

The command line was:
cargo build --target wasm32-unknown-unknown

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,24 +35,25 @@ zip = { version = "0.5.8", default-features = false, features = ["deflate"] }
 getrandom = { version = "0.2.3", features = ["js"] }
 
 # # dependencies because of need to build wasm version for file i/o
-# yew = "0.18.0"
-# wasm-bindgen = "0.2.78"
-# wasm-bindgen-futures = "0.4.28"
-# js-sys = "0.3.35"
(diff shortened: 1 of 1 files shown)


NSoiffer:
Sorry -- missed this PR for some reason.

A prior commit changed the zip compression algorithm to avoid the need to use Clang. Between that and publishing MathCAT, I believe MathCATDemo should build MathCAT without a problem now.

@jasonjgw: is your proposed change still useful?

jasonjgw:
I wasn't able to complete the compilation process for WebAssembly even with this change, but that was prior to the revision you mentioned regarding zip compression.

If it builds for WebAssembly without modification, then I'd leave this change out for the moment and close the pull request, but if you think it would be better to include those dependencies conditionally (in WebAssembly only) for the future then the approach taken is this PR could be a good starting point.

Issue #36, opened 2022-05-19
Once invalid rules dir set it is not possible to recover by setting a valid rules dir


mwhapples:
If I make a call to set_rules_dir with an invalid path, then call it again with a valid rules directory, calls to set_mathml will still give an error.
I can see a number of cases where this matters:
* Should the application let the user pick a rules directory, it would be tricky for the application to recover when the user goes on to select a valid rules directory. In some application work flows (eg. GUI applications) there should not be the need to restart just because of a change like this in a user setting.
* In testing bindings the library may be loaded once and shared amongst all tests. Should a test check that calling set_rules_dir will lead to the bindings throwing an appropriate Exception/error, this issue means no following tests will work. This is how I discovered this issue.

PR #37, opened 2022-05-19
Allow recovering from invalid call to set_rules_dir

mwhapples:
This fixes issue #36 by clearing PreferenceManager.error when initialisation succeeds. Also a test to demonstrate the issue and prove the fix works.


NSoiffer:
Looks good. Thanks for the fix!

PR #53, opened 2022-08-28
Update dependency information in callers.md

physikerwelt:
For me, the current dependency information did not work. However, with this setting, I copied from an existing extension, it worked.


NSoiffer (commented):
I changed this so it say "look for what the latest version is and use that"

PR #129, opened 2023-07-01
Convert tabs to spaces in yamls

brichwin:
Contains a python script that recursively searches for yaml files from the current directory. If a yaml file is found that has tab characters in it, then the tab character is replaced with spaces assuming tab stops every four characters. 

IMPORTANT NOTE: The python script it is checking the file as a text file, not processing the file (as a yaml file). So, it assumes that tab characters would not be coded a literal string values in the yaml file. If that is a bad assumption, then let me know.

The pull request also contains the processed yaml files starting from the Languages directory.


NSoiffer (approved):
Thanks

PR #136, opened 2023-07-24
Updated New Braille and Speech Definitions for Vietnamese Language

danghoaiphuc:
Translated some new strings for speech.
Added tan, cotan, arcsin, arccos, arctan, arccot to Braille vietnam_rules.yaml file.
Made some corrections and updated Braille definitions in unicode.yaml.


NSoiffer:
These look good but I think there are still many more changes that need to be made.

PR #160, opened 2023-08-18
update and improve speech and Braille for Vietnamese

danghoaiphuc:


NSoiffer (approved):
Looks good.

I figured out a nice way to handle the dropped numbers in a numeric fraction. I'll try to remember to add that rule on Monday.

PR #166, opened 2023-08-25
added test cases for Vietnamese speech and Braille

danghoaiphuc:


NSoiffer (approved):
Good to see some speech tests. Hopefully you can add a lot more.

PR #171, opened 2023-08-29
added Braille test cases and unit definitions for speech

danghoaiphuc:


NSoiffer (approved):
The list of units seems pretty short. Aren't there more units MathCAT should recognize?

It might be a little bit before I get to fixing up the code so the tests work.

PR #178, opened 2023-09-16
Improve Spanish for double symbols

nvdaes:
For the short styles, "abrir" plus the name of the double sing has been used.
For the detailed style, a grammatically correct form that explains what happens is used. In English maybe translated, for example, as "parentheses is opened". I don't find the similar form to right parent (paréntesis derecho) mentioned by RAE, and in fact this form has two words like abrir paréntesis, used for the short style.
Seems not to be a very defined criterion to determine differences between styles, and criterion may be difficult to define for different languages (for example, number of words, prosody, i.e., what sounds more long/short etc). So I'm using, for the first style, an abbreviate form used frequently; and for the longer styles, an explanatory construction.

### DIFF
--- a/Rules/Languages/es/unicode.yaml
+++ b/Rules/Languages/es/unicode.yaml
@@ -42,16 +42,16 @@
         then_test:
             if: "$Verbosity='Terse'"
             then: [T: "abrir paréntesis"]                	# 	(en: 'open', google translation)
-            else: [T: "paréntesis izquierdo"]   	# 	(en: 'open paren', MathPlayer: 'paréntesis', google: 'paréntesis abierto')
-        else: [T: "paréntesis izquierdo"]       	# 	(en: 'left paren', MathPlayer: 'paréntesis', google: 'paréntesis dejado')
+            else: [T: "se abren paréntesis"]   	# 	(en: 'open paren', MathPlayer: 'paréntesis', google: 'paréntesis abierto')
+        else: [T: "se abren paréntesis"]       	# 	(en: 'left paren', MathPlayer: 'paréntesis', google: 'paréntesis dejado')
  - ")":                                         	#  0x29
     - test:
         if: $SpeechStyle = 'ClearSpeak' or $SpeechStyle = 'SimpleSpeak'
         then_test:
             if: "$Verbosity='Terse'"
             then: [T: "cerrar paréntesis"]                  	# 	(en: 'close', google translation)
-            else: [T: "paréntesis derecho"]     	# 	(en: 'close paren', MathPlayer: 'cierra paréntesis', google: 'paréntesis cercana')
-        else: [T: "paréntesis derecho"]         	# 	(en: 'right paren', MathPlayer: 'cierra paréntesis')
+            else: [T: "se cierran paréntesis"]     	# 	(en: 'close paren', MathPlayer: 'cierra paréntesis', google: 'paréntesis cercana')
+        else: [T: "se cierran paréntesis"]         	# 	(en: 'right paren', MathPlayer: 'cierra paréntesis')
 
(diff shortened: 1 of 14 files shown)


NSoiffer (changes requested):
I'm confused about all the tests that you submitted. It looks like only the first two tests in functions.rs have been translated. This means that all the other tests will fail.

It Okay to have failing tests as a reminder than some functionality needs to be changed or added. However, having test failures because the tests are bad is wrong. I think you should remove the untranslated tests. You can always add them back when/if they get translated.

nvdaes:
Can please provide an example of tests that should be removed? Seems 
that I'm not understanding this. I tried to change just a few tests.

NSoiffer:
The first two tests in tests/es/SimpleSpeak/functions.rs are translated into Spanish, but the rest aren't. Here's the third test:
```
fn inverse_trig() {
    let expr = "<math><msup><mi>sin</mi><mrow><mo>-</mo><mn>1</mn></mrow></msup><mi>x</mi></math>";
    test("en", "SimpleSpeak", expr, "inverse sine of x");
```
That's English. I was a little hasty in saying it will fail. It will pass, but it isn't useful because it is a copy of an existing test. The "en" should be "es" for the Spanish tests. You did that for the first test which was translated, not for the second one that was translated. The remainder are untranslated and should either be translated (preferable) or removed (because they are duplicates of an existing test).

nvdaes:
ok thanks. i"ll try to fix it later or tomorrow when I have my computer.Enviado desde mi iPhoneEl 30 sept 2023, a las 6:53, NSoiffer ***@***.***> escribió:﻿
The first two tests in tests/es/SimpleSpeak/functions.rs are translated into Spanish, but the rest aren't. Here's the third test:
fn inverse_trig() {
    let expr = "<math><msup><mi>sin</mi><mrow><mo>-</mo><mn>1</mn></mrow></msup><mi>x</mi></math>";
    test("en", "SimpleSpeak", expr, "inverse sine of x");

That's English. I was a little hasty in saying it will fail. It will pass, but it isn't useful because it is a copy of an existing test. The "en" should be "es" for the Spanish tests. You did that for the first test which was translated, not for the second one that was translated. The remainder are untranslated and should either be translated (preferable) or removed (because they are duplicates of an existing test).

nvdaes:
I've updated tests for Spanish. Please let me know if something needs to be done.

NSoiffer (changes requested):
That's great for that file, but there are bunch of other files in your original commit that either need to be similarly modified or removed from the commit.

If that is hard to do (I'm not sure how to do it without reading some tutorial), let me know and I'll do the merge and then remove the files from the merged code.

nvdaes:
Please remove files as needed, and in the future perhaps I can do more work progressively with some guidance.

Thanks.

 

De: NSoiffer ***@***.***> 
Enviado el: miércoles, 4 de octubre de 2023 2:44
Para: NSoiffer/MathCAT ***@***.***>
CC: Noelia Ruiz Martínez ***@***.***>; Author ***@***.***>
Asunto: Re: [NSoiffer/MathCAT] Improve Spanish for double symbols (PR #178)

 

@NSoiffer requested changes on this pull request.

That's great for that file, but there are bunch of other files in your original commit that either need to be similarly modified or removed from the commit.

If that is hard to do (I'm not sure how to do it without reading some tutorial), let me know and I'll do the merge and then remove the files from the merged code.

NSoiffer:
Ok. I'll do that on Thursday. I've been very unproductive on number
localization MathML cleanups which has been my focus for the last week. I'm
close to being done... I think.

    Neil

NSoiffer:
Merging, but I'll delete the untranslated files from main.

PR #197, opened 2023-10-18
Some UEB and Nemeth symbol additions and fixes

rob-aph:
Add symbols for Degrees Fahrenheit, Degrees Celsius (Nemeth and UEB).
Fix Nemeth symbols for parallel to and not parallel to.


NSoiffer (approved):
In 8dc4361, I changed 0x2225/6 from those characters. I don't remember why. That change is clearly wrong. I'll see if any test fails.

Thanks for the fixes.

PR #198, opened 2023-10-20
Create rust.yml

GauteR:
Added GitHub workflow to automatically build and test the application as commits are done to the main branch

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,22 @@
+name: Rust
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+env:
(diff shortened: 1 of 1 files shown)


NSoiffer:
Thanks for doing this!

Some braille tests fail, but I know they fail -- I'm waiting for clarification from the people who wrote the braille manual. I'll mark them "ignore" so the test can complete.

The script looks remarkably clean and simple. I thought there was setup needed to tell it to get Rust, but I guess not.

Onto to getting it to getting the auto build to happen for MathCATForPython?

GauteR:
No problem! I can't take full credit though, I used a build template for Rust which seemed simple enough. 

I can look into the auto build for Python as well, but as that needs to produce an artifact that would have to be tagged for release, it can take some time. No promises though.

PR #200, opened 2023-10-25
UEB: Grade 1 not required for expression above and below

rob-aph:
I checked carefully, and as closely as I can tell, Grade 1 mode is not required for expression directly above and expression directly below. These changes are passing the braille UEB tests


NSoiffer (approved):
Good catch.

I looked at my tests and the only ones that would be affected by the change are examples with a sum and limit from section 7. Both of those are already in G1 mode. Too bad they didn't have an example where this made a difference.

PR #216, opened 2023-11-10
zh-TW chinese translation

hjy1210:
Make a translation of MathCat for traditional chinese(zh-TW).


NSoiffer (commented):
In looking at other people's translations, I can sometimes spot a questionable translation, but I'm at a complete loss looking at your translation to see whether an individual translation is correct or not. I do see you were able to make use of a number of the "seeded" translations. I hope that was helpful.

Overall, I'm really impressed by how thorough a translation you have done. I wish I could evaluate the quality of the speech. Do you have any questions that I can try to answer or did you feel you understood everything you needed to understand?

Just checking: leaving off the ClearSpeak rules was intentional, correct? I suspect they don't make much sense for Chinese.

Comments:
* I also see that ordinals numbers don't seem to be used in Chinese. However, you did leave one instance in "overview.yaml". Was that intentional? It's not particularly important though because I really need to do much more work to make overviews more useful.

* default.yaml: I see two rules on "mfrac" that speak the numerator before the denominator. That seems wrong.
* unicode-full.yaml: I see that you did most of the transations (very impressive!) and reverted some of them back to English (mostly in the first ~220 lines). Is there a reason for that?

Test files:
* Thank you for working on the test files -- they help me know if a change I make breaks the translation. That is useful to know when I make my change rather than finding out months later when someone complains that the MathCAT translation doesn't speak correctly (or fails completely).
* Were you able build MathCAT and run the test files using Rust? If so, do they all pass? Because I haven't pulled down your files, I don't know if Rust says there are errors or not?
* Because there is no ClearSpeak translation, remove the ClearSpeak test files -- there is no reason to keep them there. If you ever do implement ClearSpeak, you can add them back.
* As you correctly noted, some of the tests in the shared file are for ClearSpeak, some for SimpleSpeak, and some test both. For the ones that are ClearSpeak-only, assuming you don't plan to implement ClearSpeak, you should either remove the test or change "ClearSpeak" to "SimpleSpeak" and make the output correct for SimpleSpeak. I only added the ability to test both later on in the development, so the tests for whether ClearSpeak or SimpleSpeak were being tested as often based on trying to get a balance between the number of tests for each speech style.

As said earlier, overall a very impressive effort! I'll approve the PR after you let me know about the comments above.

NSoiffer:
FYI: the code has now been merged that supports a language subdomain. So when you deal with my comments, your translation should work with a new NVDA build.

hjy1210:
> I also see that ordinals numbers don't seem to be used in Chinese. However, you did leave one instance in "overview.yaml". Was that intentional?

This is my mistake. 

When I investigate why I do not find this mistake. I found all unittest DO NOT touch the rule overview::mroot. Then I found english version of unittest also DO not touch overview::mroot. 
Then I found something looks like duplicates is Speak_Rules and overview. overview::mroot is similar to SimpleSpeak::root. It seems SimpleSpeak::root have higher priority.
After inspect english version of the rule SimpleSpeak::root, may be I found some bug in there. If we perform following test, we can find the issue.
```
#[test]
fn ordinal_root_2() {
    let expr = "<math>
                    <mroot> <mi>x</mi> <mn>9.1</mn> </mroot>
                </math>";
    test("en", "SimpleSpeak", expr, "the 9.1-th root of x,");
}
```
Ther error message is
```
  left: "the 9.1-th root of x,"
 right: "the of x,"
```
I think the problem arise form Line 49-50 of SimpleSpeak_Rules.yaml
```
      - else_if: "*[2][not(contains(., '.'))]"
        then: [{x: "ToOrdinal(*[2])"}, {t: "root"}]        # phrase(the square 'root' of x)
```
lack `else part`.

I also suspect Line 78 in english version of overvie.yaml should up one level(indentation less)
```
      - x: "*[1]"
```
otherwise when `IsNode(*[1], 'simple')` is not true, nothing about `*[1]` will output.

As a summary, I will fix  my mistake.
I suggest MathCAT **can arrange an unit test to test overview::mroot(I do not know how) to benefit other translators.**

hjy1210:
> default.yaml: I see two rules on "mfrac" that speak the numerator before the denominator. That seems wrong.

The first one is correct. It is about combination not ordinary fraction. 

The second one is correct also if it is also about combination. But my unit test/integration test never touch it.
**May be you can tell me how to create a unit test about this one**.

> unicode-full.yaml: I see that you did most of the transations (very impressive!) and reverted some of them back to English (mostly in the first ~220 lines). Is there a reason for that?

Many characters in Unicode we have no chinese name. I think keep it in english is more appropriate. 
If I think some chinese name is appropriate I mark if from "t" to "T". If I am not sure I just leave it as "t".

> Were you able build MathCAT and run the test files using Rust? If so, do they all pass? 

I do run  `cargo test Languages::zh` and got following message
```
test result: ok. 244 passed; 0 failed; 1 ignored; 0 measured; 450 filtered out; finished in 2.40s
```

> Because there is no ClearSpeak translation, remove the ClearSpeak test files -- there is no reason to keep them there. If you ever do implement ClearSpeak, you can add them back.

I will remove ClearSpeak test files

> you should either remove the test or change "ClearSpeak" to "SimpleSpeak" and make the output correct for SimpleSpeak.

I need some time to do it.

Thank you for your patient.

hjy1210:
> default.yaml: I see two rules on "mfrac" that speak the numerator before the denominator. That seems wrong.

Very sorry, I am wrong. I am confused it with tag `binomial` in general.yaml.

I will correct it.

NSoiffer:
I appreciate all the work you did, especially getting the tests to pass.

NSoiffer:
I'm getting close to a release with your work in it. I'd like to credit you on my [MathCAT page](https://nsoiffer.github.io/MathCAT/#acknowledgements). Is that ok? If so, what name and what affiliation should I use?

hjy1210:
My name is 
Hon-Jang Yang

NSoiffer:
I've added your name. Take a look and see if there are any changes you would like me to make.

hjy1210:
Very Thanks

PR #219, opened 2023-11-13
More comparison operators

rob-aph:
Add Parallel To, Perpendicular to, and Up Tack to NemethComparisonOperators. Gets spaces working around these operators.


NSoiffer (commented):
Hopefully these additions are good for Nemeth also. If not, we can break the list apart. I'm planning to have a definitions.yaml for the braille rules also that could share a common file.

PR #221, opened 2023-11-14
Add White Pentagon and White Hexagon

rob-aph:


NSoiffer (commented):
Looks good. Thanks for finding and adding these.

PR #247, opened 2024-01-16
Added bridging functionallity to PHP.

kalaspuffar:
This PR will add a couple of functions to the code so you can add this library as a PHP module and run it directly from the PHP code. I've seen that other glue codes have separate repositories, so this might not be how to implement this. I am still very early in my Rust journey, so I've not looked into multi-repository solutions.

I also did a small bug fix so the library doesn't have the name `liblibmathcat.so`

Best regards
Daniel

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,6 @@ documentation = "https://nsoiffer.github.io/MathCAT/"
 edition = "2018"
 exclude = ["src/main.rs", "docs", "PythonScripts"]    # should have "Rules/", but then one can't run build.rs to build the zip file
 
-
 [features]
 "include-zip" = []
 "enable-logs" = ["android_logger"]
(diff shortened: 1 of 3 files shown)


NSoiffer:
Apologies for not seeing this earlier.

The way I have been dealing with language bindings is to have a separate repo (e.g, MathCATForC, MathCATForPython) that has wrappers for that language. The idea was not to "burden" the main library with these bindings. I don't think the current setup of using separate repos is a good idea and welcome suggestions for how to manage this better.

Two ideas that have come up:
* Make use of [Cargo Workspaces](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html) 
* Make use of github submodules

If you have any thoughts on either approach or another approach, I'd appreciate your feedback. I have heard from some people that I should avoid github submodules.

Also see https://github.com/NSoiffer/MathCAT/issues/193 for another approach to language bindings.

PR #269, opened 2024-05-29
Complete Finnish TTS with translated tests

samimaattaCelia:
Includes the Finnish translation and tests for Finnish TTS. In the shared.rs file I have added more tests that come from the Finnish matriculation exam board's official "equation book".

These changes can be merged to the main branch of MathCAT.

**Further development**
There are still some unsolved things relating to Finnish math notation, since some of the notation doesn't exist in MathML. There are also places where grammar should be improved, but it is understandable at this point. Some points on grammar:
- large operators (sum, product, integral) assume that the upper and lower bounds are numbers
- "elements belonging to a set" speech uses infinitive forms for the words, when object forms should be used, this requires new rules to implement.

### DIFF
--- a/Rules/Languages/fi/ClearSpeak_Rules.yaml
+++ b/Rules/Languages/fi/ClearSpeak_Rules.yaml
@@ -607,7 +607,7 @@
   variables: [{log_is_simple: "IsNode(*[3],'simple')"}]
   match:
   - "count(*)=3 and"
-  - "*[1][self::m:mi][text()='log' or text()='ln'] and"
+  - "*[1][self::m:mi][text()='log' or text()='ln' or text()='lg'] and"
   - "*[2][self::m:mo][text()='⁡']"
   replace:
   - test:
@@ -615,6 +615,8 @@
       then_test:
       - if: "*[1][text()='log']"
         then: [{T: "log"}]      # phrase(the 'log' of x)
+      - else_if: "*[1][text()='lg']"
+        then: [{SPELL: 'lg'}] # phrase(the 'lg' of x)
       - else_if: $ClearSpeak_Log = 'LnAsNaturalLog'
         then: [{T: "luonnollinen log"}]      # phrase(the 'natural log' of the product of 2 numbers)
         else: [{SPELL: "'ln'"}]
--- a/Rules/Languages/fi/ClearSpeak_Rules.yaml
+++ b/Rules/Languages/fi/ClearSpeak_Rules.yaml
@@ -474,10 +474,12 @@
   match: "."
   variables: [{WordToSay: "IfThenElse($ClearSpeak_AbsoluteValue = 'Kardinaliteetti', 'kardinaliteetti', 'itseisarvo')"}]
   replace:
+  # FI: not needed
   # - test:
   #     if: "$Verbosity!='Terse'"
   #     then: {T: ""}      # phrase('the' absolute value of 25)
   - x: "$WordToSay"
+  # FI: not needed
   # - T: ""      # phrase(the absolute value 'of' 25)
   - x: "*[1]"
   - test:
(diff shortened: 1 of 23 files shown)


NSoiffer:
Thanks for the contribution!

You said large operators (integrals, sums, etc) only work if the limits are numbers. Are different words used in Finish when they are variables or expressions?

Let me know if there is something I can do to help you get the rules for sets to work as you want.

NSoiffer:
Great to see a lot of tests! For me, I get one test failure after I pulled in your changes:
```
thread 'Languages::fi::ClearSpeak::mroot::mroot_above_20_simple_end_root' panicked at tests\common\mod.rs:43:23:
assertion `left == right` failed:
test with fi/ClearSpeak with prefs [
    (
        "ClearSpeak_Roots",
        "RootEnd",
    ),
] failed
  left: "21 juuri x plus y, loppu juuri;"
 right: "kaksikymmentä ensimmäinen juuri x plus y, loppu juuri;"
```

It seems like the expected answer in the test is wrong. Could you verify and fix the test?

Also, I learned how to say "plus" in  Finnish :-)

samimaattaCelia:
I remember at some point decided that indexes above 20 would be pronounced as the number itself without conjugation (i.e. "ensimmäinen"). It should just read out 21 in Finnish as it is.

I wasn't sure how I should implement it, so I left it as is. Is this something you could make a rule for? Then it should pass the test.

NSoiffer:
For roots greater than 20, I changed
```
      - else_if: "*[2][not(contains(., ','))]" # FI: Changed decimal point to decimal comma
        then: [{x: "ToOrdinal(*[2])"}, {T: "juuri"}]      # phrase(the square 'root' of 25) # FI: this needs to be changed so that when the index is bigger than 20, it doesn't use ToOrdinal() function. This already has a test "mroot_above_20_simple_end_root"
```
to be
```
      - else_if: "*[2][not(contains(., ','))]" # FI: Changed decimal point to decimal comma
        then_test:
            if: "*[2][.<=20]"
            then: [{x: "ToOrdinal(*[2])"}, {T: "juuri"}]      # phrase(the square 'root' of 25)
            else: [{x: "*[2]"}, {T: "juuri"}]      # phrase(the square 'root' of 25)
```
and now the test passes. I did this for both `ClearSpeak_Rules.yaml` and `SimpleSpeak_Rules.yaml`.

NSoiffer:
> With integral we should talk about subscript (lower bound) and superscript (upper bound) rather than use the same pattern as with other large operators.

Does this apply to all integrals (double integrals, contour integrals, etc?). You can see a list of integrals in Rules/definitions.yaml (LargeOperators). If it only applies to no more than four, then just test for them (`*[1][.='∫' or ...]`) in SharedRules/general.yaml (`tag: large-op`) or if there are many, add a new entry "Integrals" in `fi/definitions.yaml` that is similar to the `LargeOperators` definition and check them with `IsInDefinition(*[1], 'Integrals')`

samimaattaCelia:
This only applies for the single `∫` with upper and lower bounds, so I implemented that.

I noticed that I had to place the rule above "bigop-both" in SharedRules/general.yaml and also include `count(*)=3` for the match to happen correctly for just lower and upper bounds of an integral.

I also changed the tests and they passed for me.

I believe I have to do another PR, so I will do it now.

Here's the code that I implemented:

```
- name: integral-both
  tag: large-op
  match: "count(*)=3 and *[1][.='∫']"
  replace:
  - x: "*[1]"
  - pause: short
  - T: "alaraja"
  - x: "*[2]"
  - pause: short
  - T: "yläraja"
  - x: "*[3]"
  - test:
      if: "following-sibling::*"
      then:
        pause: medium
```

samimaattaCelia:
I made the new pull request #275 with the changes. It's up to date with the other changes made in this thread.

PR #276, opened 2024-08-26
Finnish ASCIIMath and tests translated

samimaattaCelia:
This is a Finnish flavor of ASCIIMath. Unicode files have been translated and there are number of new tests based on the Finnish specification.

There are bound to be need for more tests, but this is a step towards it.

No rules have been modified.


samimaattaCelia:
Point about chemistry. There is some confusion from my part how the different parts should be marked up in MathML for chemistry. For example should natrium chloride be mark up as `<mi>NaCl</mi>`, `<mi>Na</mi><mi>Cl</mi>` or even as text: `<mtext>NaCl</mtext>`?

NSoiffer:
> Point about chemistry.

It should be  `<mi>Na</mi><mi>Cl</mi>`  as there are two distinct elements. MathCAT tries to do a repair to get to that, but starting with the proper MathML avoids MathCAT having to guess. Unfortunately, most MathML generators don't get it right. That's why there are a variety of different inputs in the tests.

NSoiffer:
It looks like the ASCIIMath_Rules.yaml is a copy of the original rules. If so, or if it mostly a copy, it is better to do an "include" so that fixes in one rule set get shared by both. So the file would have just the one line:
```
- include: "../ASCIIMath/ASCIIMath_Rules.yaml"
```

If you ever need to override things, you would add the special case rules before the include.

The same applies to any other files that are mainly copies of other files.

samimaattaCelia:
I got around to this just now. For clarity the `ASCIIMath_Rules.yaml` file should be named `ASCIIMath_Rules-fi.yaml`. Is this something you want to do?

I tried running some tests, but it seemed to be that the ASCIIMath-fi needs to be linked to the tests. It also seemed that the hyphen in the name isn't working well with Rust. I'm not totally sure. I have been using "ASCIIMath-fi" in the tests as well even though there are no rules yet that work.

PR #284, opened 2024-09-12
First batch of automatic tests for Swedish braille

AndersEkl:
Hi Neil
I have done a batch of automatic tests for Swedish braille. They are currently in one rust file, called swedish.rs. Let me know if you prefer to have them in several smaller files instead.
Best regards,
Anders


NSoiffer:
Looks good! Thanks for keeping the naming style useful (i.e., pointing to where the test comes from so that if there is a failure in the future, someone can look at the spec and understand why it is the way it is).

PR #285, opened 2024-09-16
Added enable-logs feature

EL-Dol:
Added a lib named android_logger and added a feature for enabling logging on Android


NSoiffer:
Thanks for the PR.

PR #312, opened 2024-11-20
Translations for bits and bytes, and also fixes for typos and such

samimaattaCelia:
I translated the bits and bytes, and also the expected output for tests. I also fixed up some typos I found from my earlier translation and also fiexed the test output, so they should pass. At this moment I couldn't run the tests myself.

There is also a unit that doesn't have a plural form:
On the line 109 in definitions.yaml, there is a unit that has no plural form in Finnish (eng. “molar”, fin. “molaarinen”). It is used as is for plurals. This should be reflected in the rules.

### DIFF
--- a/Rules/Languages/fi/definitions.yaml
+++ b/Rules/Languages/fi/definitions.yaml
@@ -41,7 +41,7 @@
     "lm": "lumen", # lumenia
     "lx": "luks", # luksia
     "N": "newton", # newtonia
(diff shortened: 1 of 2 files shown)


NSoiffer:
> On the line 109 in definitions.yaml, there is a unit that has no plural form in Finnish (eng. “molar”, fin. “molaarinen”). It is used as is for plurals. This should be reflected in the rules.

This should be handled by adding it to the plural's list. I'll do that.

PR #321, opened 2024-12-12
translatons of units in tranditional chinese

hjy1210:
Make translatons of units in tranditional chinese and pass tests in units.rs


NSoiffer:
Thanks!

hjy1210:
Do you mean `check failed` is not an issue?

NSoiffer:
The failure is in my code. Unfortunately a change I made maybe two months ago caused a failure in one of the navigation tests. I've been meaning to improve navigation so that if an `intent` attribute is given, that changes the navigation.  That failure was the catalyst to finally make the change. It is a bigger change that I originally thought and I've been busy with some other things so I haven't gotten it all back together. Some of the failures that are listed are already fixed, but I have some more navigation fixes in those files and don't want to commit them until I get through a few more problems that need fixing.

hjy1210:
Ok, thanks for the update.

NSoiffer ***@***.***> 於 2024年12月13日 週五 下午3:35 寫道：

PR #331, opened 2025-01-05
Update users.md

NV-Codes:
+ Corrected typos
+ Updated values for various settings in line with MathCAT 0.6.6
+ Updated list of supported languages
+ Removed "Expressions are never unambiguous in SimpleSpeak."


NSoiffer:
Thanks for the fixes!

PR #338, opened 2025-01-21
Update users.md

NV-Codes:
+ Fixed typos
+ Updated information on copying math (Ctrl + C)
+ Updated list of supported braille codes


NSoiffer (commented):
Thanks for the update!

Issue #372, opened 2025-05-29
Question: Usage with a concurrent server


glebm:
I've just started looking into mathcat now and I'm wondering how one would use it from a multi-threaded / concurrent server.

The API seems to consist of global calls, so e.g. if one request calls `set_preference` with English while another thread does `set_preference` Spanish, it is unclear what the result would be.

NSoiffer:
I suppose it was maybe being lazy, but I made MathCAT single threaded since that was easier to do. MathCAT will complete whatever call you make before responding to the next call. You'll need to launch separate copies of MathCAT if you want to do concurrent things.

NSoiffer:
Thinking about it more, even if I had made MathCAT multithreaded, I don't see how that would have helped your concurrent use case. If you "randomly" call MathCAT with conflicting `set_preference` calls, what do you expect to happen? It sounds to me like you need to implement some sort of lock mechanism on your server to prevent one thread/process from interfering with what another thread/process is trying to do. Either that, or as I suggested, each thread/process should start up another instance of MathCAT.

glebm:
I expected preferences to be passed every time or held in a properties object and passed with every call.

Basically libraries generally do not have calls that modify internal global state these days.

Then, in my server, I could construct an instance of properties per language and dispatch to it.

This would work assuming that no other global mutable state is modified by the calls.

glebm:
> It would have been possible to make MathCAT stateless, but that would come at the expense of having to set up a lot of state in each call. 

In the design I'm proposing, the user can hold on to the state object, so there is no set up.
The global mutable state API can still co-exist (backed by a MathCAT-internal state instance).

PR #374, opened 2025-06-05
Speech API without global state

glebm:
This is a draft PR for #372 to demonstrate a Speech API that does not rely on mutable global state, while preserving the existing APIs that do rely on global state.

### DIFF
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -2,55 +2,18 @@
 //!
 #![allow(non_snake_case)]
 #![allow(clippy::needless_return)]
-use std::cell::RefCell;
-
+use crate::canonicalize::{as_element, name, CanonicalizeContext};
 use crate::canonicalize::{as_text, create_mathml_element};
+use crate::element_util::{get_element, trim_element};
 use crate::errors::*;
-use phf::phf_map;
-use regex::{Captures, Regex};
-use sxd_document::dom::*;
-use sxd_document::parser;
-use sxd_document::Package;
-
-use crate::canonicalize::{as_element, name};
-
+use crate::logs::enable_logs;
 use crate::navigate::*;
 use crate::pretty_print::mml_to_string;
-use crate::xpath_functions::{is_leaf, IsNode};
-
-#[cfg(feature = "enable-logs")]
-use std::sync::Once;
-#[cfg(feature = "enable-logs")]
-static INIT: Once = Once::new();
-
-fn enable_logs() {
-    #[cfg(feature = "enable-logs")]
-    INIT.call_once(||{
-        #[cfg(target_os = "android")]
-        {
-            extern crate log;
-            extern crate android_logger;
-            
-            use log::*;
-            use android_logger::*;
-        
-            android_logger::init_once(
-                Config::default()
-                .with_max_level(LevelFilter::Trace)
-                .with_tag("MathCat")
-            );    
-            trace!("Activated Android logger!");  
-        }    
-    });
-}
-
-// wrap up some common functionality between the call from 'main' and AT
-fn cleanup_mathml(mathml: Element) -> Result<Element> {
-    trim_element(mathml, false);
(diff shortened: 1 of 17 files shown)


src/interface.rs
```diff
@@ -44,6 +48,14 @@ fn enable_logs() {
     });
 }
 
+fn cleanup_mathml_with_rules<'a>(_rules: &SpeechRules, mathml: Element<'a>) -> Result<Element<'a>> {
+    // TODO: Canonicalization does not seem to actually use rules?
```
glebm, line 52: I have a question about this: I haven't been able to find where canonicalization actually uses rules, but the comment here seems to indicate it does?

https://github.com/NSoiffer/MathCAT/blob/ca4b93399aa9b3a7cbe412cac1791d45bba553f4/src/interface.rs#L109-L111

NSoiffer: Canonicalization uses the preference files (e.g., for decimal and block separators to fix up numbers). It also reads the definition files so it knows what are function names (and other things I think) for restructuring the code.

glebm:
@NSoiffer I got somewhere with this but running the example I wrote results in a strange error:

```bash
cargo build --examples
RUST_LOG=info target/debug/examples/stateless-example
```

> Error(Msg("Pattern match/replacement failure!"), State { next_error: Some(Error(Msg("attempting replacement pattern: \"default\" for \"math\".\nReplacement\n[with:\n      variables: [name: ClearSpeak_Fractions=x: \"IfThenElse($Verbosity='Verbose' and $ClearSpeak_Fractions='Auto', 'EndFrac', $ClearSpeak_Fractions)\"],[name: ClearSpeak_AbsoluteValue=x: \"IfThenElse($Verbosity='Verbose' and $ClearSpeak_AbsoluteValue='Auto', 'AbsEnd', $ClearSpeak_AbsoluteValue)\"],[name: ClearSpeak_Roots=x: \"IfThenElse($Verbosity='Verbose' and $ClearSpeak_Roots='Auto', 'RootEnd', $ClearSpeak_Roots)\"],[name: ClearSpeak_Matrix=x: \"IfThenElse($Verbosity='Verbose' and $ClearSpeak_Matrix='Auto', 'EndMatrix', $ClearSpeak_Matrix)\"],[name: MatchingPause=x: \"false()\"],[name: IsColumnSilent=x: \"false()\"],\n      replace: [test: [  if: 'x: \"$MathRate = 100\"' then:[x: \"*\"] else:[rate: 987654321.5[x: \"*\"]]]\n]]\n...due to matching the MathML\n <math id='Mo5oau70-0' data-id-added='true'>\n  <mrow id='Mo5oau70-1' data-id-added='true'>\n    <msup id='Mo5oau70-2' data-id-added='true'>\n      <mi id='Mo5oau70-3' data-id-added='true'>sin</mi>\n      <mn id='Mo5oau70-4' data-id-added='true'>2</mn>\n    </msup>\n    <mo id='Mo5oau70-5' data-id-added='true'>&#x2061;</mo>\n    <mi data-previous-space-width='0' id='Mo5oau70-6' data-id-added='true'>x</mi>\n  </mrow>\n </math>\n with the pattern\nx: \".\"\nThe patterns are in /home/gleb/MathCAT/Rules/Languages/en/SharedRules/default.yaml.\n"), State { next_error: Some(Error(Msg("Can't evaluate variable def for [name: ClearSpeak_Fractions=x: \"IfThenElse($Verbosity='Verbose' and $ClearSpeak_Fractions='Auto', 'EndFrac', $ClearSpeak_Fractions)\"] with ContextStack  0 old_values\n\n"), State { next_error: None, backtrace: InternalBacktrace { backtrace: None } })), backtrace: InternalBacktrace { backtrace: None } })), backtrace: InternalBacktrace { backtrace: None } })

I'd appreciate help figuring this one out.

glebm:
This PR has become too huge to review, so I've started sending bits of it that are not about statelessness as separate PRs with the hope of reducing the diff here.

PR #388, opened 2025-08-22
Translation crashes in Swedish rules when you have an mtable. 

kalaspuffar:
Hi Neil

I work at Textalk, and we have a close relationship with MTM and NB agencies in Sweden and Norway, and do a lot of text processing using MathCAT.

We have recently upgraded to MathCAT 0.7.0, and we got an error on one expression in the Swedish rules. Seems these have not been updated in the latest release.

The issue we get is: unknown variable  "NumColumns". We have copied the solution from the Norwegian rules, where this variable is defined.

Best regards
Daniel

### DIFF
--- a/Rules/Languages/sv/SharedRules/default.yaml
+++ b/Rules/Languages/sv/SharedRules/default.yaml
@@ -405,7 +405,9 @@
 
 - name: default
   tag: mtable
-  variables: [IsColumnSilent: false()]
+  variables:
+  - IsColumnSilent: "false()"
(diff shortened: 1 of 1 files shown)


aigan:
Minimal test case for sv:
```
<math>
  <mtable>
    <mtr>
      <mtd>
        <mrow>
          <mi>a</mi>
          <mi>b</mi>
        </mrow>
      </mtd>
    </mtr>
  </mtable>
</math>
```

PR #399, opened 2025-09-12
German translation

polx:
unicode-full.yaml in progress.
tests successful for ClearSpeak in German.
tests disabled for SimpleSpeak in German.


polx:
Trying. I've pulled the merge from main and plus-minus works. But I think it'll need some manual conflicts' resolution. Later today.

PR #403, opened 2025-09-18
Re-apply edits by @Tnonis90

CristianCantoro:
Follow-up to https://github.com/NSoiffer/MathCAT/issues/257.

I have forked the repo, recovered the edits from [commit `21ebc2e`](https://github.com/NSoiffer/MathCAT/commit/21ebc2e0d07e162b62aacbb0b64038fc4fc6fd36) made by @Tnonis90 (mentioned in the issue) and re-applied them.

### DIFF
--- a/Rules/Languages/it/SharedRules/default.yaml
+++ b/Rules/Languages/it/SharedRules/default.yaml
@@ -38,7 +38,8 @@
   match: "."
   replace:
   - bookmark: "@id"
-  - x: "translate(., $BlockSeparators, '')"   # remove digit block separators
+  # FIX: removing the digit block separators is likely locale dependent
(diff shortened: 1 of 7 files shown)


NSoiffer:
@CristianCantoro: thanks for working on this!

When you feel like the Italian translation is good enough to have other people use it, please add some tests. There are [some instructions here](https://nsoiffer.github.io/MathCAT/helpers.html#automatic-tests-for-your-translation). You don't need to copy all the tests, but please do a few files. Feel free to add `#[ignore]` to any tests that fail because you haven't done those rules yet. If you don't feel comfortable running the tests, I can run some for you and send you the output. You can then decide if it is correct; if so, you copy back to the expected output. If not, either the rules need changing or add `#[ignore]` for now.

Issue #128, opened 2023-06-27
Port MathCAT to ORCA


NSoiffer:
ORCA's math support is old and not great. ORCA is written in Python, but doesn't have an add-on architecture. MathCAT would significantly enhance math accessibility in ORCA.

If someone is part of ORCA or has contacts in ORCA, incorporating MathCAT into ORCA would be really helpful. I suspect it isn't that large a project, but that is based on just a cursory look at ORCA's code. It would likely take advantage of the [MathCATForPython ](https://github.com/NSoiffer/MathCATForPython) interface.

NV-Codes:
This would be very helpful, as Linux is often used for running simulations, in courses that teach computational/numerical techniques, and for high-performance computing, at times including the use of GUI tools for handling data, etc., in which the user uses a desktop environment (like GNOME) instead of the shell alone.

The [Orca GitHub page](https://github.com/GNOME/orca) is a read-only repository, but Joanmarie Diggs (@joanmarie) seems to be the main contributor.

It seems that [preliminary support for MathML was added to Orca in 2015](https://lists.gnome.org/archives/orca-list/2015-July/msg00013.html), but I could not find any recorded changes in Orca's handling of MathML since then. One must manually navigate with the Up and Down arrow keys to a particular line that contains a MathML expression to hear the expression verbalized; interactive navigation is not currently possible.

NSoiffer:
I had a couple of meetings with Joanmarie and two people interested in doing the ports. She suggested a way they could move forward. They went silent after the last meeting a month or two ago. I should ping them to see what's up.

PR #405, opened 2025-09-22
Improve Italian translation

CristianCantoro:
I am working on the Italian translation. The only file that still needs to be translated (completely) is `unicode-full.yaml`. This PR does the following:

  - Translate `ClearSpeak_Rules.yaml`, `unicode.yaml`
  - Revise `navigate.yaml`, `definitions.yaml`, `SimpleSpeak_Rules.yaml`
  - Revise SharedRules for Italian

I have read the ["Translator and Rule Developer Guide"](https://nsoiffer.github.io/MathCAT/helpers.html) and I am going to implement some tests in Rust. I would also like to try out the translations, but I am not an NVDA user, how can I try them out in another way? FYI, I am working on an Ubuntu Linux machine.

Also, I was wondering if the current files that I have translated are up to date. For example, I have compared the file in [`it/SharedRules/calculus.yaml`](https://github.com/CristianCantoro/MathCAT/blob/it/Rules/Languages/it/SharedRules/calculus.yaml) with the latest version of [`en/SharedRules/calculus.yaml`](https://github.com/NSoiffer/MathCAT/blob/main/Rules/Languages/en/SharedRules/calculus.yaml) and they are different.

### DIFF
--- a/Rules/Languages/it/ClearSpeak_Rules.yaml
+++ b/Rules/Languages/it/ClearSpeak_Rules.yaml
@@ -10,7 +10,7 @@
   tag: mn
   match: "starts-with(text(), '-')"
   replace:
-  - t: "meno"      # phrase(10 'minus' 4 equals 6)
+  - T: "meno"      # phrase(10 'minus' 4 equals 6)
   - x: "translate(text(), '-_', '')"
 
 - name: default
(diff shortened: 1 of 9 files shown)


NSoiffer:
Wow, you're make great progress!

I don't know when the Italian branch was created. I just looked and sadly, it is pretty far behind. I've been working on moving the "simple" rules to be declarative and listed in definitions.yaml. You can try, when you are in the "it" branch, "git merge main". You should update your fork first before doing the merge. Hopefully there aren't too many merge conflicts. If there aren't many conflicts and you can resolve them (basically accept everything that is incoming except things in the "it" branch), then update with that. After doing that, compare en/definitions.yaml with your version and bring in the new stuff. Lot's of things about units but also `IntentMappings` and `NavigationParts` definitions among other things. You'll be able to delete any of the rules listed in `IntentMappings`. You don't need to do this now if you don't want to, but at some point you'll need to do it to be up to date.

I'm hopeful I'll get help from some other people with a tool that makes is easier to keep the non-English versions synced with the English versions. Right now, it is time consuming.

As for testing the speech, running the tests is the easiest way to see if the speech is correct. Once you have something that you think is good enough, I can build a new https://nsoiffer.github.io/MathCATDemo/ that includes Italian and that's an easy way to try some things out. It's not hard to learn a few basic commands for NVDA (but it does require windows). That allows you to visit websites, etc., and listen to the math in context.

Are you a JAWS user? There might be a way to hack JAWS to use the Italian translation. I can take a look to see if it is possible (maybe by claiming the Italian is English).

NSoiffer:
Yes, both windows only. There is ORCA on Linux (see #128).  Any interest in working on that?

PR #407, opened 2025-09-24
Rebase Italian translation on top of current main

CristianCantoro:
I have [rebased](https://git-scm.com/book/it/v2/Git-Branching-Rebasing) the Italian translation on top of the current main. This means that the commits on the Italian translation files are are now applied after the latest commit on the `main` branch.

This PR compares with `main`, but I don't think it is ready to be merged into mean, so you can either [force-push](https://git-scm.com/docs/git-push) onto the `it` branch (as I have done on the [`it` branch of my fork](https://github.com/CristianCantoro/MathCAT/tree/it)) or you can create a new branch.

What remains to be done at this point is revise all the files and complete the translations that are missing.

### DIFF
--- a/Rules/Languages/it/ClearSpeak_Rules.yaml
+++ b/Rules/Languages/it/ClearSpeak_Rules.yaml
@@ -1,0 +1,784 @@
+---
+- name: intent-literal-silent
+  tag: [mi, mo, mn]
+  match: "contains(@data-intent-property, ':silent:')"
+  # say nothing
+  replace: []
+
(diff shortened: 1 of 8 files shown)


NSoiffer:
Thanks. I have updated my `it` branch with this, so I'm closing it.

Don't forget to add some tests.

FYI: I'll be on holiday in Japan for 2.5 weeks starting Friday, so I won't be able to do much. I'll be checking my mail though and should be able to answer questions if you have any. I might be slow responding though. Good luck.

PR #422, opened 2025-12-16
Switch from unmaintained `yaml-rust` to `yaml-rust2`

### LINKED ISSUE #421: `yaml-rust` is unmaintained
I recommend switching to [yaml-rust2](https://crates.io/crates/yaml-rust2).

rocallahan:
Resolves #421

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ unexpected_cfgs = { level = "warn", check-cfg = ['cfg(has_error_description_depr
 [dependencies]
 sxd-document = "0.3"
 sxd-xpath = "0.4"
-yaml-rust = "0.4"
+yaml-rust2 = "0.11"
 lazy_static = "1.4"
 strum = "0.27.2"
 strum_macros = "0.27.2"
--- a/src/definitions.rs
+++ b/src/definitions.rs
@@ -24,9 +24,9 @@
 //! See the struct [`Definitions`] for the variables that are read in.
 #![allow(clippy::needless_return)]
 
-extern crate yaml_rust;
-use yaml_rust::yaml::Hash;
-use yaml_rust::Yaml;
+extern crate yaml_rust2;
+use yaml_rust2::yaml::Hash;
+use yaml_rust2::Yaml;
 use crate::errors::*;
 use crate::prefs::*;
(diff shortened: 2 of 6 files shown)


NSoiffer (approved):
Thanks for spotting that and the changes.

NSoiffer:
@rocallahan: Hmm... this causes a failure in Rules\Languages\en\definitions.yaml, which is one of the first YAML files that is read. Did you run the tests? If so, were there no failures?  The "yaml-rust2" documentation says it is a drop in replacement for "yaml-rust", but that doesn't seem to be true.

The error "Parse error!!: invalid indentation at byte 3359 line 52 column 3" happens at the end of first definition in definitions.yaml. It didn't cause a problem in "yaml-rust" and the Red Hat YAML vscode extension thinks the file is syntactically OK. It does not appear to be a space vs tab issue as there are only spaces present in that mapping.

Do you have any suggestions. If not, I'm going to back out the change.

rocallahan:
Sorry, I didn't run the tests. You should probably back this out for now. But someone will have to fix this sooner or later I guess; RustSec has [an advisory](https://rustsec.org/advisories/RUSTSEC-2024-0320.html) marking yaml-rust as unmaintained, which triggers alarms in various CI setups.

NSoiffer:
I don't suppose you noticed anything in the documentation saying to set a certain value for compatibility with the old version? It seems strange that the new version complains when the old version didn't and the yaml syntax Checker in Visual Studio code think it's fine also.

NSoiffer:
I did some more investigation and found someone also complained about this and the author says that [the syntax is in fact invalid](https://github.com/Ethiraric/yaml-rust2/issues/30). I tried several online YAML validators and most accept it; [one ](https://yamlchecker.com/) says it is invalid. However, from what I can see reading the spec, flow sequence/mappings do not follow indentation rules. [Here ](https://dev.to/this-is-learning/yaml-collections-sequences-and-mappings-4meb#:~:text=:%20%22blue%22%20%7D-,Flow%20collection%20style,but%20empty%20node%20entries%20disallowed.&text=or%20any%20other%20white%20space%20variation.,-In%20JSON%2C%20this )and [here ](https://www.yaml.info/learn/flowstyle.html)back this up. I'll post an issue to yaml-rust2.

PR #426, opened 2025-12-23
refactor handle_special_chars() to use a single mutable buffer

moritz-gross:
the previous version creates `O(n)` temporary String objects on the heap, while this one only fills up one bigger buffer.

In general, I really like functional programming though, so maybe there is a good way to do it without doing a heap allocation for each object being mapped.

### DIFF
--- a/src/pretty_print.rs
+++ b/src/pretty_print.rs
@@ -66,20 +66,23 @@ pub fn format_attrs(attrs: &[Attribute]) -> String {
 }
 
 fn handle_special_chars(text: &str) -> String {
-    return text.chars().map(|ch|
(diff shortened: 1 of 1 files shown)


NSoiffer (approved):
Looks good. Unfortunately it won't make any real difference because AFAIK this mainly used for debugging and error messages. Still, it is good practice to be efficient unless it results in complicated code.

moritz-gross:
> Looks good. Unfortunately it won't make any real difference because AFAIK this mainly used for debugging and error messages. Still, it is good practice to be efficient unless it results in complicated code.

yeah this is what I expected. I was just looking through the codebase for the sake of practicing Rust a bit.

PR #427, opened 2025-12-25
modularize language comparison tool 

moritz-gross:
So far, everything is crammed into one big file, and doesn't use any Python packages.
This PR splits off the code by functionality, and replaces the self-written UI (e.g. the manual ANSI color codes) with the package `rich`, which replaces roughly 150 lines. No changes regarding the program logic are done yet, but I think this tool will be complex enough to need a bit more setup like this first.

So far, `PythonScripts` does not yet use a package manager, right? I personally really like [uv](https://github.com/astral-sh/uv) a lot, which has gotten really popular within a short time and is  [pip-compatible](https://docs.astral.sh/uv/pip/). 
For clarification: The current workflow is that developers just use whatever setup they like and and manually figure out which packages they might need to install? What is your preference on how to handle this?

The UI still looks very similar:
image

### DIFF
--- a/PythonScripts/README.md
+++ b/PythonScripts/README.md
@@ -1,10 +1,5 @@
 # Python Scripts
 
-This directory contains Python scripts for various uses.
-
-One of them is a tool to compare progress of localization to different languages, see the module
-`audit_translations`.
-
 Project management is done with [uv](https://docs.astral.sh/uv/).
 
 For example, execute `uv run python -m audit_translations de` to see the translation progress for the German language.


moritz-gross:
ok I'll keep it in mind for later.

I just saw that GitHub has the action "Reference in new issue" for a PR-comment. Never used it before, but sounds fitting here. Shall I?

NSoiffer:
It is probably better to just create a new issue (copying/editing the relevant parts) rather than having the irrelevant remarks, at least for this issue. But the idea that there should be an issue for this is a good one. If I don't see an issue created in a day or so, I'll hopefully remember to create one myself.

PR #434, opened 2025-12-30
fix "Fullwidth Plus Sign"

moritz-gross:
https://www.compart.com/de/unicode/U+FF0B

this symbol got messed up I think


NSoiffer (commented):
I don't think people care whether is a full width equals or not.

I'm dubious the other characters near there should say "small" also. If someone used these, it is almost certainly not meant to be different in speech/meaning and they only used them by mistake or for some visual reason.

Let me know if you feel otherwise.

moritz-gross:
yes, I also don't think this is important. The type of sign is wrong too though, it is "+" instead of "="

PR #437, opened 2026-01-02
Ru - Updates of localization

Kostenkov-2021:
This PR contains a varius changes of Russian localization at unicode symbols and ClearSpeak.yaml files. These changes are intended to clarify and correct Russian-language terms and maths phrases.


NSoiffer (approved):
I was worried that maybe you had given up on the translation. I'm very happy to see you haven't.

Is the translation now good enough for others to use?

If you feel it is ready, then if you add some tests (see instructions at https://nsoiffer.github.io/MathCAT/helpers.html#automatic-tests-for-your-translation), then I'll bring them into the main branch. Without some tests, I can't tell whether they work with the main branch and continue to work as I make changes without some tests. You do not need to do all the files (i.e., copy all the English ones and convert them to Russian), but the more you do, the higher the probability is that any problems that I might create for Russian will be caught.

PR #452, opened 2026-01-08
only show full details when using `--verbose` flag

moritz-gross:
now that I'm writing this, maybe we should call it -`-detailed`. I think "verbose" is more of a software term and might confuse people. what do you think?

default, without `--verbose`:
image

with `--verbose` (was the default until now):
image


NSoiffer:
I think the new summary output is potentially much more usable as it is clearer what to look for.

"verbose" is a pretty common option name for software, so I tend to favor using it. However, I'm not wedded to the name.

PR #458, opened 2026-01-11
Add tests for Roman numeral regex validation

moritz-gross:
while working with `LazyLock`, I saw that `UPPER_ROMAN_NUMERAL` and `LOWER_ROMAN_NUMERAL` have two matching operators `^` at the start, making them  not match leading whitespace. Is that the correct behaviour?
The comment in the code sounds like they aren't really needed generally though.

in the added test cases, the ones with leading whitespace are failing (`cargo test test_roman_numeral_regex`)

### DIFF
--- a/src/canonicalize.rs
+++ b/src/canonicalize.rs
@@ -6530,12 +6530,12 @@ mod canonicalize_tests {
 	fn test_roman_numeral_regex() {
 		assert!(UPPER_ROMAN_NUMERAL.is_match("XII"));
 		assert!(UPPER_ROMAN_NUMERAL.is_match("V  "));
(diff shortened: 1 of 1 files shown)


NSoiffer (changes requested):
When they get used, they have been trimmed. If they appear in some text string, which might include a space, then they shouldn't match.

I thought some braille codes I started with did something special for roman numerals, but they didn't so I think that is what the comment is about. However, I see at least Finnish braille cares. Looking at the code, I see several places where intended roman numerals get cleaned up (three variables V I I which should be a single token VII). Also, they are used in chemistry.

moritz-gross:
let's work on other things for now

PR #459, opened 2026-01-12
Refactor `extern crate`

### LINKED ISSUE #457: remove outdated `extern crate`
I was wondering what parts like
```
#[macro_use]
extern crate lazy_static;
```

are doing and from what I [saw online](https://www.reddit.com/r/rust/comments/ux8hb0/whats_the_use_of_extern_crate/), it's how it used to be done pre 2018,
while now the recommended way is to declare the crate in `Cargo.toml` and then just use it with `use`.

this sounds like a very quick task making the project structure a bit clearer.

if so, I'll go ahead and check this off.

moritz-gross:
`cargo check` and `cargo test` pass.

fixes #457

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -23,14 +23,6 @@ pub mod errors {
     pub use anyhow::{anyhow, bail, Error, Result, Context};
 }
 
-// Import macros for use throughout the crate
-#[allow(unused_imports)]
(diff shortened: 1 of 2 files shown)


NSoiffer:
Looks good. I can see why extern was useful (less code), but switching to a more standard "use xxx" make sense.

PR #460, opened 2026-01-13
automated translation rules for Swedish

moritz-gross:
let's keep this PR to have a nice diff for the meeting


NSoiffer:
I'll try to remember to keep this window open to show people at the meeting if it comes up.

moritz-gross:
I brought up this topic in the last DAISY metting and the general reception of using AI in the translation process was very positive/interested. 

I think it makes more sense though that when somebody want to try out using an agentic model, I'll just hold a session with that person and we'll write something from scratch, without this.

PR #461, opened 2026-01-14
get test coverage

moritz-gross:
after praising the tests today, I was thinking about how I haven't run any test coverage yet ;)

Somehow, this seems more involved for Rust than I thought it would be.
In _IntellJ_ and _PyCharm_, there are IDE-integrations that immediately let you do it.
_RustRover_ (what a name) doesn't seem to have that (yet).

I managed to get something running with [grcov](https://github.com/mozilla/grcov), which generates a report in `target/coverage/html/index.html` that looks like this:
image

- Do you already have workflow for test coverage?
- If not, what do you think about this option?
- I'm not sure if this will work on Windows nicely
- If we want to document this workflow, should we add it to the README? It does not have any technical guidance like this so far.
- Do we want to create a `CONTRIBUTING.md` for more technical information instead and add this (and some more basics) to it? I think this would be a good idea if there'll be more new people coming

Edit: I think I figured out how to do it with my IDE, but I think it's still a good idea to have a Rust-only way to do this

### DIFF
--- a/docs/developers.md
+++ b/docs/developers.md
@@ -1,0 +1,124 @@
+# Developer Guide
+
+This guide provides technical information for developers working on the MathCAT codebase.
+
+## Prerequisites
+
+To develop MathCAT, you need to have Rust installed. If you haven't already:
+
+1. [Download and install Rust](https://www.rust-lang.org/tools/install)
+2. Clone the MathCAT repository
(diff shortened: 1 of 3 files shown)


NSoiffer:
It's a start. Thanks.

PR #463, opened 2026-01-15
Add unit tests for TTS

moritz-gross:
`tts.rs` so far has no own tests, and is only indirectly covered from the other files.
Here are some basic ones to get started.
`cargo test` passes.

It took me a few minutes to get that TTS _None_ means plain-text if I understand it correctly? Should we add a comment for this in the enum? 
This is not the `None`/`Null` built into the language itself, but is just an arbitrary name inside of the enum, right? I'm even a bit surprised this is allowedin Rust, iirc Java doesn't allow a class to be named "class" for example.


moritz-gross:
ok nice.
do we want to document what _None_ means in this context another time, or do you think this is clear as it is?

PR #464, opened 2026-01-15
Add unit tests for `pretty_print.rs`

moritz-gross:
56.79 % of lines were covered indirectly, but no tests within the directly yet.

I couldn't find out what `sxd` in `sxd_document` stands for. Something related to XML I assume?

### DIFF
--- a/src/pretty_print.rs
+++ b/src/pretty_print.rs
@@ -643,4 +643,70 @@ mod tests {
             _ => panic!("Expected string value from xpath"),
         }
     }
+
+    #[test]
+    /// Evaluates non-BMP literal text with a MathML namespace-qualified XPath.
+    fn xpath_non_bmp_namespace_literal() {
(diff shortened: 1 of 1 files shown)


NSoiffer:
I have no idea of what "sxd" stands for either (the same author has "sxd-xpath" which MathCAT uses).

Can you add a test for a 6 byte character like `𝞪` (0x1d7aa) and `\u1d7aa`). I think either or both sxd-doc or sxd-xpath fail for one of those -- I've had to work around them. If it is a failure, don't include it in the test, but if it works, include it. We should create a simple test for that and submit a bug report to the crate's developer and maybe even a PR if it isn't too hard to track down (I suspect

NSoiffer:
Maybe its the YAML reader. In the last month, I ran into it. I'll need to look for where it is. In any case, it is good to have a test with these characters.

PR #472, opened 2026-01-28
rough prototype for rule coverage

moritz-gross:
as you mentioned earlier, it'd be interesting to have some form of coverage report on the rules.
I added logic for this to `src/rule_coverage.rs`, which gets used in `speech.rs` if the Cargo feature `rule-coverage` is on.
The coverage is saved to the format provided by _LCOV_, which is what I already explored previously for the vanilla test coverage.

With `cargo test --features rule-coverage`, the LCOV files can be created, and `genhtml --flat target/rule-coverage/lcov.info -o target/rule-coverage/html-flat` then creates a report in HTML format that can be viewed.

Right now, it looks pretty empty (maybe something is missing?):
image

I spent only limited effort on validating the mutex-logic and export to LCOV (which I've never used before), but I think the general idea of using a Cargo feature and then piping the results into existing coverage tooling is a good approach.

We can discuss and refine this later of course, I just couldn't help myself but try to hack this together :)


NSoiffer (commented):
Finally got around to looking at this. It looks ok, but I don't see why you are using guards. MathCAT is single threaded. Maybe I missed it, but I don't see anything here that starts up MathCAT in parallel (multiple threads), so that seems like extra complication.

moritz-gross:
> Finally got around to looking at this. It looks ok, but I don't see why you are using guards. MathCAT is single threaded. Maybe I missed it, but I don't see anything here that starts up MathCAT in parallel (multiple threads), so that seems like extra complication.

neither do I know. this is only a very rough proof of concept, as I wanted to see if I can get this to work in the first place.

do you think the general approach makes sense? i can take a few hours to properly read into this and make a proper PR.

NSoiffer:
Yes, I think it makes sense.

moritz-gross:
@NSoiffer When I was working on simplifying this, I noticed that `cargo test` is multi-threaded by default (even if MathCAT itself is not). It's mentioned surprisingly early and explicitly in the Rust book: https://doc.rust-lang.org/book/ch11-02-running-tests.html#running-tests-in-parallel-or-consecutively. 

So we either have to handle this somehow (I'm not a big concurrency guy, so I need some time to judge if the status quo is an adequate solution), as the results file for the tests is a shared state, or we need to use `--test-threads=1`.

My assumption would be that keeping cargo test multi-threaded is the easiest, as otherwise, we have to explain to users what `--test-threads=1` is about, and it's gonna be slower (the speedup is basically the whole point of it, right?). Also, the complexity can be absorbed by rule_coverage.rs pretty neatly.

NSoiffer:
This is marked as a draft. Is it ready for merging?

moritz-gross:
status: I want to properly learn how concurrency in Rust works before merging this, so it might take some more time.

PR #488, opened 2026-02-14
Fix output ordering

moritz-gross:
Marthe requested if we could order by line-number instead of alphabetically, which I think makes a lot of sense.

---

side note: we could add `uv run pytest` (called inside _PythonScripts_) to CI.

how does _GitHub Actions_ work with dependency tools like `cargo`/`uv`? Is `cargo build` run from scratch every time before testing? looks like it to me (https://github.com/daisy/MathCAT/actions/runs/22022276872/job/63633257352?pr=488). should we look into caching some of this to not run into any limits?

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -1,0 +1,18 @@
+name: Python
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+  test:
+    name: Python Tests
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: astral-sh/setup-uv@v6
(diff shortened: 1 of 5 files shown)


moritz-gross:
image

and there now is a new symbol in the "files changed" overview. 🤔
looks like it worked.

NSoiffer:
I'm not savvy with GitHub either. My guess is that you should be able to tell locally whether there is a problem by making sure your files are up to date.

moritz-gross:
> I'm not savvy with GitHub either. My guess is that you should be able to tell locally whether there is a problem by making sure your files are up to date.

hm ok. just did that, this PR is also good to go.

and good to see that the python tests finished in 9 seconds.
running them locally takes less than 1 second, so the vast majority is setup then.

PR #495, opened 2026-02-18
add AGENTS.md

moritz-gross:
for experimenting with agentic coding, it would be interesting to have a shared file for steering them.

afaik, `AGENTS.md` is the [convention](https://agents.md/) used by _OpenAI Codex_, as well as _Cursor_, while _Claude_ has CLAUDE.md, but I think we can gently point to this file in that case :)
in the end it's just a markdown file anyway.
I've seen projects that really overdo this, but I think having one extra file is worth it.

I added some of the mistakes I had to deal with in my experiments, like having the whole codebase nuked by `cargo fmt`, or having packages installed by `pip` when `uv` should be used instead, or having to reintroduce the `audit-translations` tool again and again.

how I imagine it, once we see any repeated mistake, we just add the solution here, and in a few weeks everything works smoothly (lets see)

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -1,10 +1,6 @@
 name: Python
 
-on:
-  push:
-    branches: [ "main" ]
-  pull_request:
-    branches: [ "main" ]
+on: [push, pull_request]
 
 jobs:
   test:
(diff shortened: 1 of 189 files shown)


NSoiffer:
Good idea.

PR #506, opened 2026-02-27
Fix chemistry test `merge_bug_274`

masonium:
This test is currently failing on the `main` branch.

I'm only somewhat confident that this is the correct fix. It does seem intentional that the underlying `canonicalize` call transforms the `<mtext>2</mtext>` tag into a number representation `<mn ...>2</mn>`.


NSoiffer:
That's correct.  I added the fix but forgot to run the tests. The next morning my (new) machine was dead :-(. So I've been bogged down with trying to fix it myself then doing the same thing with a tech support person...

Thanks for the fix

PR #508, opened 2026-02-27
run CI on all branches, not only "main"

moritz-gross:
from my understanding, `on` is an obligatory attribute (https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#on).

Also, `pre-release` is restricted to _main_ on job-level. should we do the same for `zip-rules`?

For Python CI, there's only the one job for testing, so that file looks finished.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -1,10 +1,6 @@
 name: Python
 
-on:
-  push:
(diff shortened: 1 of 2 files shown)


NSoiffer:
Thanks

PR #509, opened 2026-03-01
Add Finnish translation of user guide with language switcher

moritz-gross:
I prototyped internationalization of the docs for Finnish. The English version still lives directly in `docs`, but we now also support other langues in subfolders like `fi`, `de`, etc.

@samimaattaCelia below is your translation running in the same template that's already in use, except that I added a language switcher to it. Is this what you have in mind?
Also, I don't know what you technically mean by "considering different screen readers". Can't we just have one `user.md` file that has different sections for this?

image

@NSoiffer I ran jekyll using a mounted Docker container. How did you do local development? I haven't found a gemfile and also some entries could be added to the `.gitignore`.

Generally, how does hosting the docs work? afaik, there is not GitHub Actions workflow. does it rely on the repo settings and a hardcoded path (so `docs` I guess?)

As we already have some translations, I wanted to get this started quickly so that you are not blocked. 
let me know which parts of this prototype can stay or have to be changed. I haven't worked with Jekyll yet, so I'm very much improvising here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,8 @@ src/speech-bak
 
 Rules/*/*/*.zip
 .vs
+
+# local caches when building Jekyll
+docs/_site
+docs/.jekyll-cache
+docs/Gemfile.lock
--- a/docs/Gemfile
+++ b/docs/Gemfile
@@ -1,0 +1,3 @@
+source "https://rubygems.org"
+gem "github-pages", group: :jekyll_plugins
+gem "webrick" # HTTP server removed from Ruby 3.0+ stdlib; needed for `jekyll serve`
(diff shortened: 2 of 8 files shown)


samimaattaCelia:
Thanks, @moritz-gross, that looks nice!

>Also, I don't know what you technically mean by "considering different screen readers". Can't we just have one user.md file that has different sections for this?

Now that I read the guide again, I only mention a NVDA specific button once, so it shouldn't be a problem. I don't have access to JAWS, so somebody else would have to write if there is something that is specific to JAWS.

Best scenario would be that the users could translate even the user guide. There is one other person I know who is actively translating NVDA into Finnish using NVDA's translation track. I don't know about other add-ons and how their users guides are translated.

Issue #528, opened 2026-03-25
Letters incorrectly read as Roman numerals


MartheGjelstad:
I have noticed multiple cases where letters are incorrecly read as Roman numerals. Below are some examples and the output from MathCAT.

`<math>
    <mi>C</mi><mo>=</mo><mi>D</mi>
</math>`

Output: "100 is equal to 500"

`<math>
<mi>C</mi><mo>+</mo><mi>I</mi><mo>+</mo><mi>X</mi>
</math>`

Output: "100 plus 1 plus 10"

`<math>
<mi>I</mi>
    <mo>=</mo>
    <mo>−</mo>
    <mi>b</mi>
    <mi>r</mi>
    <mo>+</mo>
    <msup>
        <mi>z</mi>
        <mi>I</mi>
    </msup>
</math>`

Output: "cap i is equal to, negative b r, plus z to the first power"

PR #540, opened 2026-03-31
Fix Roman numeral handling #528

moritz-gross:
I added the examples from the issue as regression tests.
Generally it looks like the detection of Roman numerals should be more conservative.

tbh I'm not really happy how this seems to be handled within MathML, but I guess we have to work with what we're given.

The tests should be ready to go as they are, but I doubt the code changes are ideal.
I only just learned that there's an intent attribute, and that chemistry apparently uses roman numerals for oxidation numbers?!? (well at least that's one way to get some more general knowledge, yesterday I learned what a katal is)

### DIFF
--- a/src/canonicalize.rs
+++ b/src/canonicalize.rs
@@ -838,7 +838,7 @@ impl CanonicalizeContext {
 			},
 			"mi" => {
 				let text = as_text(mathml);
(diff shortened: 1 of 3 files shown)


NSoiffer:
The real problem was a bad Rule that missed testing for chemistry.

The roman numerals shouldn't be spoken, but some braille codes handle them in a special way (typically modifying the capitalization rules).

I've included the tests you wrong in a separate commit.

PR #543, opened 2026-04-02
fix --rules-dir in mathml2text CLI

moritz-gross:
While testing out the mathml2text CLI, I noticed that `--rules-dir` seems to not actually be used.
For example, the following shouldn't work right? `cargo run --bin mathml2text -- --rules-dir /some/other/rules`

Regarding the integration tests, I don't really know what I'm doing there.


masonium:
@NSoiffer asked me to take a look. The actual `rules_dir` fix looks good. I'll leave a couple of other comments inline.

tests/mathml2text.rs
```diff
@@ -0,0 +1,97 @@
...
+    let _ = std::fs::remove_file(&input_file);
+
+    assert!(output.status.success(), "stderr: {}", String::from_utf8_lossy(&output.stderr));
+    assert_eq!("4\n", String::from_utf8_lossy(&output.stdout));
+}
+
+/// Verifies that the positional input-file argument still works on its own.
+/// This keeps the default CLI file-input path covered while testing the rules-dir changes separately.
+#[test]
+fn still_accepts_input_file() {
```
masonium, line 48: I don't think this test is really necessary. In practice I don't think you're really testing anything besides the argument parsing library itself.

tests/mathml2text.rs
```diff
@@ -0,0 +1,97 @@
...
+        .duration_since(UNIX_EPOCH)
+        .expect("system time should be after UNIX_EPOCH")
+        .as_nanos();
+    std::env::temp_dir().join(format!("mathml2text-{name}-{timestamp}.mml"))
+}
+
+/// Verifies that `mathml2text` uses an explicitly provided `--rules-dir`.
+/// This protects the CLI override path instead of silently falling back to the default rules location.
+#[test]
+fn accepts_explicit_rules_dir() {
```
masonium, line 22: I would put most of this boiler plate code in a fixture structure, so you can reuse more of the logic.

NSoiffer:
@masonium: Should I close this without merging?

moritz-gross:
I think the comments by @masonium make sense, I just forgot that this PR existed. I'll do it in the coming days.

side note: I'm unsure how I just update this branch. 99% sure I clicked "Sync fork" in my repo, so https://github.com/moritz-gross/MathCAT should've been affected. or what am I missing?

moritz-gross:
@masonium is this what you had in mind? so far I've not written any fixtures for CLI or file system interaction, particularly in Rust. 

Is the `Drop` trait needed when using `temp_dir`? From how I unterstood the docs, it is, but I'm usure.

failure of linting in `main.rs `seems unrelated to my changes.

NSoiffer:
@masonium: where are we on this? @moritz-gross asked for some guidance and it looks like that's where this issue stands.

PR #549, opened 2026-04-18
AI seeding for French localization

moritz-gross:
This PR was generated using Codex (the coding agent from OpenAI) with these instructions:

  - Expand the French localization for MathCAT.
  - Keep the scope bounded so a native French speaker can validate it in roughly 10 hours at most.
  - Use the existing localizations as the basis, with English as the primary reference and other languages as secondary references where helpful.
  - You may use the CLI during development.
  - You may add unit tests in the style of the existing language tests.
  - Mark all AI-generated translations with lowercase `t:` instead of `T:`, since they are not yet human-validated.
  - Mark all AI-generated unit tests with `AI generated`.
  - First make a plan for that bounded French expansion, then continue from the plan and start implementing it.

---

I think it makes sense to first explore how this workflow turns out in practice, and to not have Codex burn through tokens for hours. @NSoiffer these instructions make sense to you?

Sadly I don't remember the name of the French-speaking dev from the last DAISY meeting for tagging here, but if you're reading this, hi and let me know if this is of any help and what you think of this.

### DIFF
--- a/Rules/Languages/fr/SharedRules/calculus.yaml
+++ b/Rules/Languages/fr/SharedRules/calculus.yaml
@@ -1,0 +1,56 @@
+---
+
+- name: laplacian
+  tag: laplacian
+  match: "count(*) <= 1"
+  replace:
+  - t: "laplacien"
+  - test:
+      if: "count(*) = 1"
+      then:
+      - test:
+          if: "$Verbosity!='Terse'"
+          then: [t: "de"]
+      - test:
+          if: "not(IsNode(*[1], 'leaf'))"
+          then: [pause: short]
+      - x: "*[1]"
+
+- name: divergence
+  tag: divergence
+  match: "count(*) = 1"
+  replace:
+  - test:
+      if: "$Verbosity='Terse'"
+      then: [t: "div"]
+      else: [t: "divergence de"]
+  - test:
+      if: "not(IsNode(*[1], 'leaf'))"
+      then: [pause: short]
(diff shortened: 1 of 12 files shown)


NSoiffer:
Here are the changes I would make:

  - ~~Expand the French localization for MathCAT.~~
  - Create a new fr branch -- _maybe do this by hand first_
  - Copy Rules/Languages/en to Rules/Languages/fr. This is where rules for the French language live. _maybe do this by hand first_
  - ~~Keep the scope bounded so a native French speaker can validate it in roughly 10 hours at most.~~ I'm not sure what that means
  - ~~Use the existing localizations as the basis, with English as the primary reference and other languages as secondary references where helpful.~~
  - Using the existing localizations as the basis, with English as the primary reference and other languages as secondary references where helpful, create a French translation.
  - Make sure to localize to use French math idioms and not just copy the English rules.
  - Add any rules for idioms used in French that are not part of English
  - You may use the CLI during development.
  - ~~You may add unit tests in the style of the existing language tests.~~  This is problematic -- essentially it would have AI test what AI created. That doesn't seem reliable.
  - Mark all AI-generated translations with lowercase `t:` instead of `T:`, since they are not yet human-validated.
  - Mark all AI-generated unit tests with `AI generated`.
  - ~~First make a plan for that bounded French expansion, then continue from the plan and start implementing it.~~ In cursor, you choose what mode you are in.  I would think that is an option in whatever environment you use to generate the code.

PR #550, opened 2026-04-19
Change `isize` to `i32` 

moritz-gross:
I was looking to learn Rust a bit more and saw that `isize` is used a lot in `chemistry.rs`, though I don't see why it shouldn't be a standard integer type like `i32`.
as I understand it, `isize` is primarily for pointers, ie platform dependent, which does not make sense for stuff like the `likely_chem_superscript` scoring function.

side note: I find the RegEx's like `MULTIPLE_PLUS_OR_MINUS_OR_DOT` very hard to read. What do you think about having some intermediate parts to have it in smaller chunks?


NSoiffer:
That looks good. I suspect at some point, I started with a `usize`, realized I needed negative numbers, so just went with `isize`. `i32` makes more sense. Efficiency-wise (memory/speed), it is irrelevant.

There are probably some `usize` usages that might make sense to be `u32`, but again, I'm pretty sure I used `u32` in all the spots that matter.

Issue #551, opened 2026-04-19
bug in some assertions in `chemistry.rs`


moritz-gross:
Some assertions in `chemistry.rs` have a `!` in the front, which I assume should be logical negation.
But as `.len()` returns a value of type `usize`, the `!` operator is the bitwise flip.
And as `usize` is the *unsigned* type, it's basically impossible to land at the value 0 or 1 using bitwise flipping.

Is this some kind of bug in a bug situtation, as to why it has not shown up yet?

I guess the chemistry section is not really a priority right now anyway, I was just poking around in the file as I recently talked to a friend doing his Chemistry PhD, so I was interested in what I could find here ;)

Here is one such example from the code:

```
fn is_structural(elements: &[&str]) -> bool {
    assert!(!elements.len() > 1);   // already handled
```

PR #552, opened 2026-04-20
fix assertions in chemistry.rs

moritz-gross:
see #551


NSoiffer:
Surprising that this wasn't trigged by any of the tests.

PR #558, opened 2026-04-22
allow more whitespace before audit-ignore comment

moritz-gross:
see #559


moritz-gross:
failure of Clippy/Lint check is in `navigate.rs`, so unrelated.

PR #560, opened 2026-04-22
kickoff French localization

moritz-gross:
hi @olivierthiffault, I started off with letting Codex (the AI coding tool by OpenAI) do some work and create labeled unit tests for review. let me know what you think of this and if this of any help.
If you have any feedback, we can incorporate it when we do further localization work on MathCAT.

### DIFF
--- a/Rules/Languages/fr/SharedRules/calculus.yaml
+++ b/Rules/Languages/fr/SharedRules/calculus.yaml
@@ -1,0 +1,56 @@
+---
+
+- name: laplacian
+  tag: laplacian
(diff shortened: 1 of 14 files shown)


NSoiffer:
Don't forget to add definitions.yaml

olivierthiffault:
Thank you @moritz-gross . I had a look and it seems like a good starting point. Looking at the tests there are a  a number of traductions that need correction but we will work on that

NSoiffer:
Merged into `fr` branch.

PR #563, opened 2026-04-23
make some Rust snippets more idiomatic

moritz-gross:
I want to keep practicing Rust some more, so I tried to find spots that could be refactored to use Rust language features a bit better.
not urgent of course, but overall I think it's good to do some housekeeping on the side.

### DIFF
--- a/src/interface.rs
+++ b/src/interface.rs
@@ -258,9 +258,8 @@ pub fn get_preference(name: impl AsRef<str>) -> Result<String> {
             }
             if value == NO_PREFERENCE {
                 bail!("No preference named '{}'", name);
-            } else {
-                return Ok(value);
             }
+            Ok(value)
         })
     }));
     return report_any_panic(result);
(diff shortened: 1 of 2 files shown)


src/interface.rs
```diff
@@ -258,9 +258,8 @@ pub fn get_preference(name: impl AsRef<str>) -> Result<String> {
             }
             if value == NO_PREFERENCE {
                 bail!("No preference named '{}'", name);
-            } else {
-                return Ok(value);
             }
+            Ok(value)
```
NSoiffer: I know it isn't needed, but I find adding "return" makes the code more readable to me because it makes the return value more obvious. So I prefer `return Ok(value)`.

moritz-gross: hm, I personally really like the (close-to) functional style, but I think having it more explicit is also reasonable.
I'll change it.

moritz-gross: done.

NSoiffer:
Thanks. AI or clippy pointed out some of those kinds of changes, but I obviously didn't get to all of them.

PR #574, opened 2026-05-01
handle canonicalize failure in containerized environments

### LINKED ISSUE #516: canonicalize() calls fail in restrictive environments (ACCESS_DENIED)
### Description:
When running MathCAT in restrictive environments (e.g., container), the calls to std::fs::canonicalize() fail with ACCESS_DENIED errors, preventing MathCAT from initializing. 

### Problem:
canonicalize() requires the OS to resolve symlinks and validate every component of the path, which involves filesystem-level permission checks. In environments with restricted filesystem access, this fails even when the target files are readable.

The affected locations are:
prefs.rs — PreferenceManager::initialize() (line ~263)
shim_filesystem.rs — canonicalize_shim() and read_to_string_shim()

### Expected behavior
MathCAT should gracefully handle canonicalize failures by either 
1. falling back to constructing an absolute path manually (e.g., joining relative paths with the current working directory) or 
2. by providing new api like set_rules_dir_with_absolute_path(). I guess, the core logic only needs consistent absolute paths — full symlink resolution is not required.

For the internal testing, I have tried the first approach, it is working in restrictive environments.

**prefs.rs:**

```
 pub fn initialize(&mut self, rules_dir: PathBuf) -> Result<()> {
        #[cfg(not(feature = "include-zip"))]
        let rules_dir = match rules_dir.canonicalize() {
            Err(_e) => {
                // fallback code to absolute path - Start
                if rules_dir.is_absolute() {
                    rules_dir
                } else {
                    std::env::current_dir()
                        .unwrap_or_default()
                        .join(&rules_dir)
                }
                // fallback code to absolute path - end
            },
            Ok(rules_dir) =>  rules_dir,
        };

```

**shim_filesystem.rs**

```
        pub fn canonicalize_shim(path: &Path) -> std::io::Result<PathBuf> {
            match path.canonicalize() {
                Ok(p) => Ok(p),
                Err(_) => {
                    // Fallback: if canonicalize fails (e.g., ACCESS_DENIED),
                    // construct an absolute path manually
                    if path.is_absolute() {
                        Ok(path.to_path_buf())
                    } else {
                        Ok(std::env::current_dir().unwrap_or_default().join(path))
                    }
                }
            }
        }
       
        pub fn read_to_string_shim(path: &Path) -> Result<String> {
            let path = match path.canonicalize() {
                Ok(path) => path,
                Err(_) => path.to_path_buf(), // fallback option
            };
            info!("Reading file '{}'", &path.display());
            match std::fs::read_to_string(&path) {
                Ok(str) => return Ok(str),
                Err(e) => bail!("Read error while trying to read {}: {}", &path.display(), e),
            }
        } 
```

If anyone has any other better solutions to this problem, please share your suggestions.

venkatesh6114:
### Description

Fixes https://github.com/daisy/MathCAT/issues/516

path.canonicalize() can fail with ACCESS_DENIED in containerized or sandboxed environments. This caused MathCAT to
bail out entirely when loading rules or reading files.

### Changes

 - canonicalize_shim (shim_filesystem.rs): Instead of hard-failing, falls back to returning the path as-is (if 
absolute) or prepending cwd (if relative).
 - read_to_string_shim (shim_filesystem.rs): Replaces the hard bail!() on canonicalize failure with a fallback to 
the original path, letting read_to_string attempt the read anyway.
 - PreferenceManager::initialize (prefs.rs): Same fallback logic applied to rules_dir canonicalization.

### DIFF
--- a/src/prefs.rs
+++ b/src/prefs.rs
@@ -259,9 +259,23 @@ impl PreferenceManager {
     /// 
     /// If rules_dir is an empty PathBuf, the existing rules_dir is used (an error if it doesn't exist)
     pub fn initialize(&mut self, rules_dir: PathBuf) -> Result<()> {
+        // Resolve the rules directory to an absolute, canonical path.
+        // If canonicalize() fails (e.g., ACCESS_DENIED in containers), fall back to:
+        //   - returning the path as-is if it is already absolute,
+        //   - prepending the current working directory if it is relative.
+        // Note: if current_dir() also fails, unwrap_or_default yields an empty PathBuf,
+        //       and the result may remain relative.
         #[cfg(not(feature = "include-zip"))]
         let rules_dir = match rules_dir.canonicalize() {
-            Err(e) => bail!("set_rules_dir: could not canonicalize path {}: {}", rules_dir.display(), e),
+            Err(_e) => {
+                if rules_dir.is_absolute() {
+                    rules_dir
+                } else {
+                    std::env::current_dir()
+                        .unwrap_or_default()
+                        .join(&rules_dir)
+                }
+            },
             Ok(rules_dir) =>  rules_dir,
         };
 
--- a/src/shim_filesystem.rs
+++ b/src/shim_filesystem.rs
@@ -340,14 +340,31 @@ cfg_if! {
             }
         }
         
+        /// Resolves the path to an absolute, canonical form using the OS.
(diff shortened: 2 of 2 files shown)


venkatesh6114:
@NSoiffer Could you please review this. I appreciate if you have any other suggestions as well.

NSoiffer:
Sorry for taking so long to review this.

PR #586, opened 2026-06-01
Addition of Russian documentation

Kostenkov-2021:
This commit adds Russian documentation for users, developers and translators.


NSoiffer:
I'm very impressed that you are implementing Russian braille! I'd very much like to see at least 100 examples that cover different parts of the braille system. I couldn't find any online spec, but I did see that [this Wikipedia page](https://en.wikipedia.org/wiki/Russian_Braille) has three examples. I added them to `russian.rs`. All of them fail. I don't trust that the examples are correct, but even if they aren't correct, the output has non-braille characters and that's clearly wrong. Here's an example:
```
---- braille::Russian::russian::wikipedia_linear_parens stdout ----
  left: "⠼⠉⠄⠣⠼⠊⠀⠤⠼⠛⠜⠀⠶⠼⠋"
 right: "⠼⠉'\\⠭⠼⠚⠚⠚⠚⠃⠛'(⠼⠊⠀⠤⠼⠛)⠀⠶⠼⠋"' at tests\common\mod.rs:152:0
```

I'll bring over the documentation into main so it should show up on the web page.

Kostenkov-2021:
Hello. Thank you for adding documentation. I will continue working on braille. thank you for your feedback.

PR #589, opened 2026-06-03
Enhance Russian Braille support with mathematical symbols and rules

Kostenkov-2021:
This PR adds fixes and expands Russian Braille support for MathCAT.


NSoiffer (commented):
There is a much better way to handle the long if...elseif...elseif....

I was surprised I didn't see any example in a braille file, but there is in an english file. In definitions.yaml, you want to add something like:
```
- SIPrefixes: {
    "Q": "quetta", "R": "ronna", "Y": "yotta", "Z": "zetta", "E": "exa", "P": "peta", "T": "tera", "G": "giga", "M": "mega", "k": "kilo", "h": "hecto", "da": "deka",
    "d": "deci", "c": "centi", "m": "milli", "µ": "micro", "n": "nano", "p": "pico", "f": "femto", "a": "atto", "z": "zepto", "y": "yocto", "r": "ronto", "q": "quecto"
  }
```
but in this case, you want the braille.  So something like
```
- TrigFunctions: {
}   "arcsin": "⠫⠁⠎#", ...
```
and in the rule file you use
```
"DefinitionValue(. 'Braille', 'TrigFunctions')
```

FYI: unless Russian braille is very simple, I think you will need emulate some of the other braille files and use "L" for letters, "N" for digits, and a few other things so that after the rules are applied, you can do some cleanup. For example, if there is a rule that a letter after a digit is allowed accept for a-j which are used for digits for 0-9 in Western braille codes, you can adjust the code in the final cleanup. Without the "L" and "N", there is no way to detect that case.

NSoiffer:
There are conflicts. I started to resolve them by accepting all the incoming changes, but when I got to definitions.yaml, that seemed like the wrong thing to do. Can you please resovle the conflicts as I don't know what is correct.

NSoiffer:
Thanks for the fixes.

Any idea when you think the braille will be good enough for inclusion into the main branch?  Right now it looks like you have 59 tests. There should be over 100 tests, and preferably 200 tests that cover a wide range of notations. In braille, other codes have lots of rules about grouping, so if grouping/not-grouping is something that is relevent to Russian braille, then make sure you have plenty of tests for that (e.g, superscripts with 'x', '-x', '-2', 'x_2', nested superscripts, supscripts followed by letters or numbers, or whatever the tricky cases are.

PR #592, opened 2026-06-04
Core concept names additions and tests -- issue 381

### LINKED ISSUE #381: Add the rest of the core concept names to MathCAT
The [core concepts names](https://w3c.github.io/mathml-docs/intent-core-concepts/#functions-and-inverses) is in a nearly final state. 

For each language, there is a definitions file that (should) specify how to speak that intent.  For example, in `Rules/Languages\en\definitions.yaml` there is the `IntentMappings` variable. For any intent that fits a simple generic pattern, the intent should be listed for that variable. For example,
```
    "absolute-value": "function= ; absolute value: the absolute value: the absolute value; end absolute value",
```
Some of the translations don't yet have this variable in the file.

Some concepts have special cases (e.g., fractions, powers), and so they need an actual rule.

The values in 'en' and some other directories were added as a test and are probably only ~20% of the concept names. The rest need to be added, and for non-English files, eventually translated.

For those concepts that don't fit within what `IntentMappings` allows, a note should be made that a rule needs to be added.

For translations, "t:" and "T:" indicates whether something is translated. That doesn't work here. Probably the only thing to do is to add a comment
```
# NEEDS TRANSLATION
```
to the end of the line.

ttpyorre:
Issue: https://github.com/daisy/MathCAT/issues/381

Added the rest of the core concept names to Rules/Languages/en/definitions.yaml from the site: https://w3c.github.io/mathml-docs/intent-core-concepts/

Tests have been created in tests/Languages/en/definitions.rs. Also added comments for each definition that might be broken either on the site, or has weird behavior when tested.


NSoiffer (commented):
Not sure why the nav test fails. I'll look into it.

PR #596, opened 2026-06-08
fix rules for digits in german

moritz-gross:
Me and Codex went through the German localization once again for a few minutes to spot easy fixes.
This one is kinda funny as in English, bold has two meanings as in "brave" or "thick", which doesn't exist in German.

side-question: why is so much of `alphabet.rs` in the *de* tests commented out?


NSoiffer:
Looks good. 

FYI:  I have someone who is working for me that I have tasked to use ranges for these chars. Hopefully he is going to do a PR soon -- it should be an easy project using AI

PR #605, opened 2026-06-22
Update of Swedish translation

TimArborealis:
YAML files have been updated as per the present date. Existing rust tests have been updated as needed to pass but new ones have not been added and translated. Navigation has not been tested, awaiting the creation of automatic tests.

Also fixed a few minor typos in English unicode-full.


NSoiffer:
Looks like this is ready to merge into main. Is that correct?

PR #612, opened 2026-06-26
Braille unicode range fixes

ttpyorre:
Fixed some Nemeth and UEB ranges. 

Added Swedish tests to the test branch so they get recognized, however tests ex_6_15, ex_6_16, and ex_6_19 are not yet working. These tests seemed to have been broken before the unicode-full.yaml changes from https://github.com/daisy/MathCAT/pull/606. Before merging the branch, should see if we can fix them in this branch, or if we should fix them separately and add them to another issue.


ttpyorre:
Swedish tests under tests/braille/Swedish/swedish.rs are breaking, there are three

ex_6_15
  left: "⠃⠣⠼⠚⠀⠲⠀⠿⠦⠁⠣⠼⠁⠳⠃⠣⠼⠁⠲⠿⠦⠁⠣⠼⠃⠳⠃⠣⠼⠃⠲⠀⠄⠄⠄⠀⠲⠁⠣⠝⠳⠃⠣⠝⠿⠴⠿⠴"
 right: "⠃⠣⠼⠚⠀⠲⠀⠿⠦⠁⠣⠼⠁⠳⠳⠃⠣⠼⠁⠲⠿⠦⠁⠣⠼⠃⠳⠳⠃⠣⠼⠃⠲⠀⠄⠄⠄⠀⠲⠁⠣⠝⠳⠃⠣⠝⠿⠴⠿⠴"' 
left check has ⠳⠃instead of the output of  ⠳⠳⠃at two parts.

ex_6_16
  left: "⠵⠀⠶⠀⠿⠦⠼⠛⠭⠤⠼⠋⠽⠳⠼⠑⠭⠲⠼⠓⠽⠿⠴⠳⠳⠿⠦⠼⠛⠭⠤⠼⠋⠽⠳⠼⠓⠽⠲⠼⠑⠭⠿⠴"
 right: "⠵⠀⠶⠀⠿⠦⠿⠦⠼⠛⠭⠤⠼⠋⠽⠳⠼⠑⠭⠲⠼⠓⠽⠿⠴⠳⠳⠿⠦⠼⠛⠭⠤⠼⠋⠽⠳⠼⠓⠽⠲⠼⠑⠭⠿⠴⠿⠴"' 
left has ⠿⠴ instead of the output of ⠿⠴⠿⠴ 

For the two tests above, it seems to me the left side is wrong, and need to be corrected in the test, but I will leave it for someone who is better at braille.

and for ex_6_19
  left: "⠿⠦⠼⠊⠳⠼⠋⠳⠳⠼⠉⠿⠴⠂⠀⠿⠦⠼⠊⠳⠼⠋⠌⠌⠼⠉⠿⠴⠀⠕⠉⠓⠀⠿⠦⠼⠊⠌⠼⠋⠳⠳⠼⠉⠿⠴⠀⠃⠑⠞⠑⠉⠅⠝⠁⠗⠀⠞⠁⠇⠑⠞⠀⠼⠁⠂⠑⠳⠼⠉"
 right: "⠿⠦⠼⠊⠳⠼⠋⠳⠳⠼⠉⠿⠴⠂⠀⠼⠊⠳⠼⠋⠌⠌⠼⠉⠀⠕⠉⠓⠀⠿⠦⠼⠊⠌⠼⠋⠳⠳⠼⠉⠿⠴⠀⠃⠑⠞⠑⠉⠅⠝⠁⠗⠀⠞⠁⠇⠑⠞⠀⠼⠁⠂⠑⠳⠼⠉"'
Not entirely sure what all is going on in this test.

PR #616, opened 2026-06-29
Ru - Add braille label marks and tests

Kostenkov-2021:
Introduce comprehensive Russian braille label handling: new label/mark rules (exact-over/under, upper/lower-right, repeated/no-indicator, left-labels) and mmultiscripts prescript/postscript logic in Rules/Braille/Russian/Russian_Rules.yaml. Add Unicode braille mappings and new definition maps (RussianLabelMarks, RussianExactOverLabelMarksWithoutIndicator, RussianUpperRightLabelMarksWithoutIndicator) in definitions.yaml and unicode.yaml, and change '%' mapping. Expand tests (tests/braille/Russian/russian.rs) with cases for percent/special marks, labeled table rows, mmultiscripts variants, digit sequences, infinity index, complex sup fractions and many label-mark expectations.


NSoiffer:
Thanks for continuing to work on this.

Can you explain why you added `propagate_inherited_mathvariant` and the related `propagate_mathvariant_to_children` in canonicalize.rs?[ `mathvariant` is only allowed on leaf elements](https://www.w3.org/TR/mathml4/#presm_commatt), so it should never be inherited.

Also, `normalize_math_digits` seems unneccessary and not in the correct location anyway. `merge_block` is called when it looks there is a block of digits (`is_likely_a_number`). So any test needs to be earlier. However, the test `is_likely_a_number` uses regular expressions that contain `\d`, and in Rust, that includes all Unicode digits.

Are there some examples/bugs where this isn't working?

Kostenkov-2021:
Hello @NSoiffer 
`propagate_inherited_mathvariant` was added as a workaround for a few Russian braille tests that used `<mstyle mathvariant="...">` to represent a styled span. I removed both propagation helpers and changed the Russian braille tests to use leaf-level `mathvariant` instead.

I also removed `normalize_math_digits`. `merge_block` is already downstream of `is_likely_a_number`, and the regexes using `\d` already recognize Unicode digits. The existing `digit_block_non_ascii_int` and `parent_bug_94` tests still pass without that normalization, so I don't currently have a remaining example that requires it.

NSoiffer:
Thanks. How much more work do you think you have before the Russian braille is ready to go? I see 30 tests (all passing). Typically I'd like to see around 200 tests, but at least 100.

Note: in braille.rs, I'm now using traits instead of a "match" for things like `russian_cleanup`. I modified the code accordingly. You should update your branch to match what is in github (which updated to main).

Kostenkov-2021:
Hello @NSoiffer 
I agree the current test count is misleading/too low from the CI perspective. The Russian braille file currently has 30 Rust test functions, but many of them contain multiple examples; locally I count about 240 individual `test_braille` checks. Still, I understand that the test output only shows 30, so I’ll split the larger grouped tests into smaller named `#[test]` cases and add coverage where it is still thin, aiming for at least 100 visible tests and preferably closer to 200.

For readiness, I think the main remaining work is:
1. sync the branch with current main and make sure the new trait-based braille dispatch is matched;
2. split/expand the Russian braille tests so CI reports enough individual tests;
3. do one more pass over the Russian source material for missing edge cases;
4. run full `cargo test` and `cargo clippy --all-targets -- -D warnings`.

Locally, `braille.rs` already appears to use the new `BrailleCode` trait dispatch, including `impl BrailleCode for Russian`, so I expect that part to be mostly alignment rather than a rewrite.



> Thanks. How much more work do you think you have before the Russian braille is ready to go? I see 30 tests (all passing). Typically I'd like to see around 200 tests, but at least 100.
> 
> Note: in braille.rs, I'm now using traits instead of a "match" for things like `russian_cleanup`. I modified the code accordingly. You should update your branch to match what is in github (which updated to main).

Issue #415, opened 2025-12-05
Navigation issue on <msubsup>


Lucylinxchang0102:
It can’t navigate into `<msubsup>` completely, for example: 
Given the MathML: `<math><msubsup><mi>𝑥</mi><mn>1</mn><mn>2</mn></msubsup></math>`
And navigate by below commands:
1. Zoom in --> “in base; [[x sub 1”
2. Next --> “end of math”
3. Zoom in --> “in base; [[x sub 1”
4. Next --> “end of math”
	
The result always skips “x”, “subscript 1”, “superscript 2” when doing the navigation.

michaldziwisz:
Confirming this still reproduces on `0.7.6-beta.5`, and it is language-independent (identical behavior in English and in other locales, so it's clearly in the core navigation logic, not in any language rules).

Exact MathML from the report:
```xml
<math><msubsup><mi>x</mi><mn>1</mn><mn>2</mn></msubsup></math>
```
Enhanced mode, ClearSpeak:
1. ZoomIn -> "in base; x indexed by 1"  (base `x` and subscript `1` are already glued together -- you cannot land on `x` or on `1` individually)
2. MoveNext -> "in exponent; 2"
3. MoveNext -> "end of math"

So the base and the subscript are never individually reachable -- the deepest you can get is the combined "x indexed by 1", and everything to the left of the exponent is skipped.

The same skipping happens for a plain `<msub>` embedded in a larger `<mrow>`. For `y = x_1 + x_2`:
```xml
<math><mrow><mi>y</mi><mo>=</mo><msub><mi>x</mi><mn>1</mn></msub><mo>+</mo><msub><mi>x</mi><mn>2</mn></msub></mrow></math>
```
1. ZoomIn -> "y"
2. MoveNext -> "is equal to"
3. MoveNext -> "x indexed by 1; plus x indexed by 2"  (the whole `x_1 + x_2` is read as one block; MoveNext never steps into the individual subscripted terms)
4. MoveNext -> "end of math"

A standalone `<msub>` behaves the same way (ZoomIn reaches "in base; x", Next reaches the subscript, but the base `x` and subscript cannot be visited as separate leaves).

Happy to provide additional minimal cases if useful.

MartheGjelstad:
Documenting my testing with the same example. Used verbosity verbose for the navigation.
Used NVDA 2026.1.1 and the built-in MathCAT.
[...]

PR #623, opened 2026-07-01
Add Polish (pl) language localization

michaldziwisz:
This PR adds a Polish localization for MathCAT's speech and navigation rules.

What's included (all under `Rules/Languages/pl/`, 13 rule files):
- ClearSpeak_Rules.yaml, SimpleSpeak_Rules.yaml
- SharedRules/ (default, general, geometry, linear-algebra, calculus)
- definitions.yaml, unicode.yaml, unicode-full.yaml, intent.yaml, navigate.yaml, overview.yaml

Plus regression tests:
- tests/Languages/pl.rs -- Polish rule regression tests
- tests/languages.rs -- registers the pl test module

Notes:
- No changes to the engine core (src/ untouched).
- Written against the crate version currently on main (0.7.6-beta.5).
- The rules have been extensively tested against corpora generated via pandoc and against real-world MathML from Wikipedia (Mathoid). Continuous reading is correct across all tested domains; the remaining known navigation limitations for a handful of nested-norm (U+2225) and degree (degC) cases are language-independent core-navigation behaviors (see e.g. #415), not specific to these rules.

Happy to adjust anything based on your review.

### DIFF
--- a/Rules/Languages/pl/ClearSpeak_Rules.yaml
+++ b/Rules/Languages/pl/ClearSpeak_Rules.yaml
@@ -1,0 +1,803 @@
+---
+# Polska wymowa operatorow arg max / arg min / ess sup. Operator (mo, mi lub mrow[arg,max])
+# jest emitowany przez x: w regule named-under-operator (SharedRules/default.yaml) - emisja
+# czyni go nawigowalnym; tutaj nadajemy mu wlasciwa nazwe PL zamiast doslownego "arg max".
+- name: op-arg-max-min-ess-sup
+  tag: [mo, mi, mrow]
+  match:
+    - "( (self::m:mo or self::m:mi) or"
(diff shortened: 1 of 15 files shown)


NSoiffer:
Thanks for getting this done!

I moved the test file pl.rs into a "pl" directory to be consistent with how the other tests work. 

Will you have time to add more tests? There are currently 56 Polish tests; most of the other languages have over 500 tests (copied from the "en" tests).

michaldziwisz:
Follow-up: I've opened #641 with the additional tests you asked for — Polish is now at 605 tests (up from 56), mirrored from the English suite with expected speech captured from the actual engine output. It also fixes Polish plural declension in multi-line equations ("2 równania" / "5 równań" instead of the English-style "2 równanies"). All changes are limited to tests/Languages/pl and Rules/Languages/pl; src/ is untouched.

PR #640, opened 2026-07-28
fix: distinguish principal value Log in speech

### LINKED ISSUE #637: Multi-valued logarithms
_From Mohammed Laachir_

We are looking at a JAWS customer report and wanted to ask whether this is likely a MathCAT issue, a missing semantic markup case, or expected behavior.

The customer is reading this Complex Analysis page:

https://complexanalysis.org/web/sec_complex-log.html

The page distinguishes the multivalued complex logarithm, written as lowercase log(z), from the principal value, written as uppercase Log(z). The customer's concern is that JAWS reads both forms the same way, even though general log has multiple values while principal Log has one principal value.

From our side, JAWS gets the speech for math from MathCAT. I also checked the current MathCAT repo on GitHub:

- README: MathCAT converts MathML to speech strings and is used by JAWS.
- Rules/Languages/en/definitions.yaml maps "logarithm" to function=log and "natural-logarithm" to natural log, but I did not see a separate principal-logarithm mapping.
- tests/Languages/en/SimpleSpeak/functions.rs has tests for lowercase <mi>log</mi>, expecting "the log of x" / "the log of, open paren x plus y, close paren". I did not see an analogous test for uppercase <mi>Log</mi> or principal logarithm.

Could you let us know how you think MathCAT should handle this?

Specifically:

1. If the MathML contains <mi>Log</mi>, should MathCAT infer and speak it as "principal log" or "principal logarithm" in English?
2. Or would MathCAT only be expected to say "principal logarithm" if the source MathML includes an explicit intent, such as a future/appropriate principal-logarithm intent?
3. If this is a MathCAT bug or missing rule/test case, is it something you would prefer we file upstream in MathCAT with the page and example MathML?

Example distinction from the customer page:

- log(z): multivalued complex logarithm
- Log(z): principal value of the logarithm, ln|z| + i Arg(z)

The customer-facing issue is that a speech user cannot tell which version is present when reading the page in JAWS.

w3lld1:
## Summary

- I added English speech handling for uppercase `Log` as the principal-value logarithm.
- I added regression coverage for both SimpleSpeak and ClearSpeak.
- Lowercase `log` behavior remains unchanged.

## Validation

- `cargo test Languages::en::SimpleSpeak::functions -- --nocapture` (35 passed)
- `cargo test Languages::en::ClearSpeak::functions -- --nocapture` (49 passed)
- `git diff --check upstream/main...HEAD`

I left the full repository build and test suite to CI to keep local validation focused on the changed English speech path.

Fixes #637

### DIFF
--- a/Rules/Languages/en/SharedRules/general.yaml
+++ b/Rules/Languages/en/SharedRules/general.yaml
@@ -465,9 +465,9 @@
       else: [t: "covariance"]      # phrase('covariance' function)
 
 -
-  name: log    # handle both log and ln (if in an mrow, 'intents' are used)
+  name: log    # handle log, principal Log, and ln (if in an mrow, 'intents' are used)
   tag: mi
-  match: ".='log' or .='ln'"
+  match: ".='log' or .='Log' or .='ln'"
   replace:
   - bookmark: "@id"
   - test:
@@ -476,6 +476,8 @@
   - test:
     - if: ".= 'log'"
       then: [t: "log"]      # phrase(the 'log' function is used in mathematics)
+    - else_if: ".= 'Log'"
+      then: [t: "principal value log"]
     - else_if: "$Verbosity='Terse'"
       then: [spell: "'ln'"]
       else: [t: "natural log"]      # phrase(the 'natural log' function is used in mathematics)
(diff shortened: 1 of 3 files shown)


NSoiffer:
Looks good. Thanks for the fix.

Issue #14, opened 2022-03-21
Add to check for ending punctuation


NSoiffer:
There is currently a check to avoid speaking punctuation at the end of a math expression.

[This Wikipedia page](https://en.wikipedia.org/wiki/Probability) has a `.` followed by a `<mspace with='thinmathspace'/>` and that is _not_ being caught.

Make canonicalization throw out spaces at the start/end (if in `math`/`mrow`).

NSoiffer:
Note: there is a test for this called `ignore_period_and_space`. It is currently has `#[ignore]`.

PR #641, opened 2026-07-31
Expand Polish test suite (56 -> 605) and fix plural declension in multi-line equations

michaldziwisz:
Follow-up to #623. After the Polish language pack was merged you asked whether I could add more tests (Polish had 56, most languages have 500+). This brings Polish to **605 tests**.

## What's here

**1. Tests mirrored from the English suite (56 -> 599).**
Structure mirrors `tests/Languages/en/` (ClearSpeak, SimpleSpeak, shared, definitions, units, chemistry, alphabets, mtable). Expected speech strings were **captured from the actual engine output** with the Polish rules and then verified as correct Polish, rather than translated by hand.

A few cases keep engine-level behavior that is identical in English and Polish (an intent without explicit fixity; the `root` intent without an index, which errors in both languages) — these are annotated in-place, and one such tuple is skipped rather than asserting a core-level quirk.

**2. Fix: Polish plural declension in multi-line equations (599 -> 605).**
Polish declines a counted noun by number: 1 -> singular, 2-4 (except 12-14) -> nominative plural, otherwise genitive plural. The multi-line rules were copied from English and appended an `s` suffix, producing wrong forms like *"2 równanies"* or *"5 równania"*. Fixed in both speech paths:
- `SharedRules/general.yaml` (SimpleSpeak)
- `ClearSpeak_Rules.yaml` (ClearSpeak — all labels: Case/Line/Constraint/Equation/Row/Step)

Now: **1 równanie / 2 równania / 5 równań** (and 12 równań, 22 równania). Polish-specific tests cover singular, paucal, genitive plural and the 12-14 exception in both styles.

## Scope
All changes are limited to `tests/Languages/pl/` and `Rules/Languages/pl/`. **Engine core (`src/`) is untouched.** `cargo test --test languages Languages::pl` -> 605 passed, 0 failed, 1 ignored (the ignored one mirrors en's `#[ignore] // issue #14`).

### DIFF
--- a/Rules/Languages/pl/ClearSpeak_Rules.yaml
+++ b/Rules/Languages/pl/ClearSpeak_Rules.yaml
@@ -631,36 +631,59 @@
   - LineCount: "count(*[not(contains(@data-intent-property, ':continued-row:'))])"
   - NextLineIsContinuedRow: "false()"   # default value
   - IsColumnSilent: true()
+  # Polish plural declension category of $LineCount: singular (1), paucal (2-4
+  # except 12-14 -> nominative plural), otherwise genitive plural.
+  - Paucal: "($LineCount mod 10 = 2 or $LineCount mod 10 = 3 or $LineCount mod 10 = 4) and not($LineCount mod 100 = 12 or $LineCount mod 100 = 13 or $LineCount mod 100 = 14)"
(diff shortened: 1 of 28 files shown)


NSoiffer:
Thanks for the speedy additon! I'll merge this into main.

I'll update this branch -- you should continue to make PR's to 'pl', not to main.

PR #644, opened 2026-08-03
fixed missing translation in General.yaml

Tnonis90:
Pull requesting the translation fix for the Diagonal rule


NSoiffer:
FYI: I'm planning to do a release soon. As part of that process, I'm going to do a create a release candidate branch next week. If you feel the Italian translation is close to ready, then if you can get a bunch of tests done, I'll include it in the release.

I hope to do releases roughly every month. NVDA will pick them up roughly every three months. I don't know how often JAWS will pick them up. Because I am (once again) going to be away for a month starting in late August, the next release will be more like 6 weeks after the upcoming release.

Issue #617, opened 2026-06-29
Support definitions.yaml


moritz-gross:
@MartheGjelstad  @TimArborealis (and more people I think)

I've not yet started working on supporting `definitions.yaml` files, mostly as I've not yet worked with them and don't really know what this is about.

From what I can see, the `definitions.yaml` files have a different structure than "normal" rule files, correct? Is their structure documented somewhere that I should look into? Also, are there even more file-types the audit-translations tool should support, that we should keep in mind? Should I specify in the tool that this only applies to definitions.yaml, or is it possible that this also applies to files with a different name? Does this introduce any changes to how the CLI is used, like having a new flag like `--check-defintions` or alike?

Basically, I don't think this whole thing is very hard from a programming perspective, the main thing that's needed for me here is to understand the status quo and goal behaviour better.

TimArborealis:
The audit tool seems to catch all the files which need updating except definitions.yaml, so I don't think there are any more file types to consider. The Swedish update is complete now and definitions was the only file we had to do "manually".

MartheGjelstad:
In `navigate.yaml` there are instances of 

- "out of"
- "in"
- "end of"
- "start of"

than should be translated. These are not captured by the audit-tool. 

They always come after the word "Move2D" like this: 

`Move2D: "'in'"`

Also, in the rule `into-or-out-of-mtr`, the match needs to be translated. The audit-tool now captures this as a diff.

NSoiffer:
Unlike the other yaml files, there is nothing in definitions.yaml that says whether something has been translated or not. Maybe something should be added or maybe the audit tool makes that mute?

moritz-gross:
@NSoiffer how should I treat this in the data-modeling? let's take this shortened example (from `Rules/Languages/en/definitions.yaml`:

```
- PluralForms: {
    "degree celsius": "degrees celsius",
    "mile per gallon": "miles per gallon",
  }
```

So far, I have a dataclass `RuleInfo` for the normal rule files, which assumes the fields `name` and `tag` (among others). This contract is broken here, so does it make sense to have a separate "definitions-rule" dataclass?

moritz-gross:
@MartheGjelstad @TimArborealis I added functionality to the tool

One thing that's lacking is an analogous thing like `# audit-ignore`, but for definitions, however we might want to use this in practice. I've not considered this for now and only compare the files as they are.

I think this is a good step forward and I'll add further behaviour on request.

[...]

PR #647, opened 2026-08-06
Exclude `definitions.yaml` from audits

moritz-gross:
before implementing the ideal behaviour, let's at least mute it for now.

see issue #617


moritz-gross:
@NSoiffer fuzzer is running for 45 mins now. is this expected behaviour?

image

Edit: it was 142 minutes in the end.

PR #660, opened 2026-08-09
simplify display of RulesFor variants

moritz-gross:
a small simplification using the strum crate that we already have.

### DIFF
--- a/src/speech.rs
+++ b/src/speech.rs
@@ -16,6 +16,7 @@ use sxd_xpath_no_unsafe::{Factory, Value, XPath};
 use sxd_xpath_no_unsafe::nodeset::Node;
 use std::fmt;
 use std::time::SystemTime;
+use strum_macros::Display;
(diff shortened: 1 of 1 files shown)


NSoiffer:
Looks good. I'm surprised that there aren't other places that can take advantage of that macros (or that it is not built-in).

moritz-gross:
> Looks good. I'm surprised that there aren't other places that can take advantage of that macros (or that it is not built-in).

there are multiple places that are close this, but not exactly fitting. 

For example fmt of `Token` in `infer_intent.rs` is a close candidate, but we have the special case of
`Token::ConceptOrLiteral(str) => format!("Literal({str})"),`, which breaks the clean variant-based formatting.

Another case is "sapi5" for `TTS::SAPI5` in `prefs.rs`, which switches from upper- to lowercase.


I could just make PRs based on what looks the cleanest and most consistent to me and then you can judge if it should be changed ?

PR #661, opened 2026-08-09
Ru - Improve wording of Russian docs

Kostenkov-2021:
Replace "Трудности в обучении" with "Когнитивные нарушения" in docs/ru/User_guide_for_MathCAT_ru.md to use a more precise and respectful term for users who need shorter spoken output.


NSoiffer:
I'm not sure what commands you are using for git, but you should only be commiting your changes, not all the others. Except for possibily src/braille.rs, all of your PR commits should be restricted to Russian rules and tests.

FYI: the ru branch is updated to main (but more updates to main are coming).

PR #666, opened 2026-08-11
use strum_macros::Display for Token

moritz-gross:
ok I forgot that you also need to reuse `str` in the display-string, so this makes it a bit longer than I thought it would be.

still, that's a 9 line improvement, and it keeps the logic closer by having it directly next to each variant, instead of an extra function.

in my book, this is still worth doing


NSoiffer (commented):
I'm not that happy with this change. I think it makes reading the enum much harder for a minor drop in the total number of lines.

Do you want to provide a counter argument?

PR #701, opened 2026-08-18
Add option --exclude to audit-translations

CristianCantoro:
Hello, I am going over the [instructions](https://daisy.github.io/MathCAT/helpers.html) for validating new translations and I have found useful to add an `--exclude` option to `audit-translations`.

This option allow excluding specific files from the audit. In a sense, it does the opposite of what `--file` does, but `--file` can be used to specify only a single file.

I have added the following examples in the `README.md`:
```sh
# Exclude a list of files from the audit, note that if you use this option before specifying the
# target language you'll need to use the option terminator (--)
uv audit-translations es --exclude unicode-full.yaml
uv audit-translations es --exclude unicode-full.yaml unicode.yaml
uv audit-translations --exclude unicode-full.yaml -- es
```

### DIFF
--- a/PythonScripts/audit_translations/README.md
+++ b/PythonScripts/audit_translations/README.md
@@ -77,6 +77,7 @@ uv run --project PythonScripts audit-translations --list
   * Region variants are shown as `lang-region` (e.g., `zz-aa`) based on subdirectories under `Rules/Languages/<lang>`.
 * `--source`: Sets the source/reference language. Defaults to `en`.
 * `--file`: Audits a single specific file instead of the whole directory.
+* `--exclude`: Exclude one or more files from the audit.
 * `--rules-dir`: Override the Rules/Languages directory path.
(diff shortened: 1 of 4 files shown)


PythonScripts/audit_translations/README.md
```diff
@@ -101,8 +102,15 @@ uv run audit-translations de
...
 
 # Audit only a specific file
+uv run audit-translations es --file ClearSpeak_Rules.yaml
 uv run audit-translations es --file SharedRules/default.yaml
 
+# Exclude a list of files from the audit, note that if you use this option before specifying the
+# target language you'll need to use the option terminator (--)
+uv audit-translations es --exclude unicode-full.yaml
+uv audit-translations es --exclude unicode-full.yaml unicode.yaml
+uv audit-translations --exclude unicode-full.yaml -- es
```
moritz-gross: should be `uv run audit-translations` instead of `uv audit-translations`

moritz-gross (approved):

PR #712, opened 2026-08-22
Add Simplified Chinese speech and navigation rules

cary-rowen:
## Summary

This PR adds a complete Simplified Chinese translation for MathCAT's speech and navigation rules.

The current English rules were used as the source, but the spoken text was rewritten in natural Mainland Mandarin rather than translated word for word. Terminology follows common usage in Mainland Chinese mathematics education.

The rules are placed directly under `Rules/Languages/zh`, allowing `zh-CN` and similar locales to use them through MathCAT's normal locale fallback. The existing Traditional Chinese overrides under `Rules/Languages/zh/tw` remain unchanged.

## Coverage

The translation includes:

- ClearSpeak, SimpleSpeak, and LiteralSpeak output
- arithmetic, calculus, geometry, and linear algebra rules
- fractions, roots, powers, matrices, sets, intervals, coordinates, units, and common chemistry notation
- mathematical intents and common and extended Unicode symbols
- overview descriptions and mathematical navigation
- locale fallback and Simplified Chinese regression tests

All 12 required language rule files are included, and verified translations use MathCAT's `T:` marker.

## Translation choices

The wording is kept concise and follows forms familiar to students and screen reader users in Mainland China.

- Fractions use denominator-first order, such as `2 分之 1`.
- Powers and roots use natural Chinese word order, such as `x 的 n 次方` and `x 的 n 次方根`.
- Set and logic notation uses established readings such as `空集`, `属于`, `减去`, `任意`, `且`, `或`, and `推出`.
- Geometry notation uses semantic readings such as `三角形`, `平行四边形`, `相似于`, `全等于`, `直角三角形`, and `圆 O`.
- `dot-product(a,b)` is read as `a 与 b 的数量积`, while `cross-product(a,b)` is read as `a 与 b 的叉积`. Ordinary infix operators remain `点乘` and `叉乘`.
- Independent or prefix `±` and `∓` are read as `正或负` and `负或正`; binary operations are read as `加或减` and `减或加`.
- Variants of the less-than-or-equal and greater-than-or-equal signs use `小于等于` and `大于等于` instead of visual descriptions of the glyphs.

## Context-dependent symbols

Some symbols have more than one mathematical meaning and are handled according to context.

A bare `∆` is read as `增量`, while an expression with the explicit `laplacian` intent is read as `拉普拉斯算子`. Likewise, `⊙O` is read as `圆 O`, while a binary expression such as `a ⊙ b` keeps the generic reading `带圈点运算符`.

Symbols such as `∼` and `□` are not assigned a single specialized meaning where the presentation alone is insufficient to determine their use.

## References

Terminology and symbol readings were checked against:

- [GB/T 3102.11-1993《物理科学和技术中使用的数学符号》](https://openstd.samr.gov.cn/bzgk/gb/newGbInfo?hcno=3DE79450D562E62D41CB6E79FF411054)
- 人民教育出版社《普通高中课程标准实验教科书 数学（B版）选修2-1》
- [全国科学技术名词审定委员会“术语在线”](https://www.termonline.cn/)
- [Unicode Character Database](https://www.unicode.org/Public/17.0.0/ucd/UnicodeData.txt)

## Validation

The Chinese language test suite passes with 376 tests and no failures. The translation audit covers all 12 rule files and reports no untranslated strings or definition errors. Its two apparent missing-rule findings are grouping differences where the Chinese rules split combined English character ranges to preserve distinct readings. The completed diff also passes whitespace validation.

## Scope

This PR only adds the base Simplified Chinese rules and their tests. It does not change the existing Traditional Chinese rules, braille rules, or MathCAT engine code.


NSoiffer (commented):
Thanks for for doing this!

It looks like you used AI to generate this. Have you looked through the tests and/or the unicode files to make sure the AI made the appropriate translantion? At least for other languages, it struggles to get the math terminolgy correct. For the tests, it tends to just make the expected results match the computed ones.

FYI: I changed the merge target to daisy:zh-cn (which I updated to merge).

PR #758, opened 2026-09-08
ci: add scheduled workflow to alert translators of English rule changes

### LINKED ISSUE #496: Alert Translators on English File Changes
To to ease the work for translators and to assure the quality of non-canonical languages over time, translators need to be notified when the English files are updated.

**Possible Solution:**
Introduce a new GitHub Actions workflow that automatically notifies translators whenever changes are pushed to monitored English‑language files. The workflow posts a new discussion (to not clutter issues) that links to relevant commits, and tags translators as a team or individually.

Johannes-spsm:
Closes #496

### DIFF
--- a/.github/translator-alert.config
+++ b/.github/translator-alert.config
@@ -1,0 +1,7 @@
+# Add your GitHub username (including '@') to the 'notify' list to receive alerts when English rules change.
+# Append to the existing list on the same line, separated by a single space (e.g. notify=@user1 @user2).
+# You do not have to add your Github username if you are already in the mathcat-translators group.
+
+paths=Rules/Languages/en/
+exclude=Rules/Languages/en/gb/
+notify=@johannes-spsm @daisy/mathcat-translators
--- a/.github/workflows/translator-alert.yml
+++ b/.github/workflows/translator-alert.yml
@@ -1,0 +1,268 @@
+name: Alert Translators on English File Changes
+
+permissions:
+  contents: read
+  discussions: write
+
+on:
+  schedule:
+    - cron: '3 14 1,15 * *'
+
+jobs:
+  alert:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+
+      - name: Load alert config
+        id: config
+        run: |
+          CONFIG_FILE=".github/translator-alert.config"
+          
+          PATHS=$(grep '^paths=' "$CONFIG_FILE"   | cut -d '=' -f2-)
+          EXCLUDE=$(grep '^exclude=' "$CONFIG_FILE" | cut -d '=' -f2-)
+          NOTIFY=$(grep '^notify=' "$CONFIG_FILE" | cut -d '=' -f2-)
+          
+          echo "paths=$PATHS"    >> $GITHUB_OUTPUT
+          echo "exclude=$EXCLUDE" >> $GITHUB_OUTPUT
+          echo "notify=$NOTIFY"  >> $GITHUB_OUTPUT
+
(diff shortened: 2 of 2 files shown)


.github/translator-alert.config
```diff
@@ -0,0 +1,7 @@
+# Add your GitHub username (including '@') to the 'notify' list to receive alerts when English rules change.
```
moritz-gross: maybe add "in this file" or "see below" or something like this.
Based on the first sentence alone, I first thought you'd have to edit `translator-alert.yml`

moritz-gross: using mathcat-translators here is really neat I think, hadn't thought of this before.

.github/workflows/translator-alert.yml
```diff
@@ -0,0 +1,268 @@
+name: Alert Translators on English File Changes
+
+permissions:
+  contents: read
+  discussions: write
```
moritz-gross: could this be scoped down so that write-access is only for the announcements? This is what the discussions are are used for anyway, and it would make it easier to see what this permission is used for.

.github/workflows/translator-alert.yml
```diff
@@ -0,0 +1,268 @@
...
+
+jobs:
+  alert:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
```
moritz-gross: I've never worked with fetch-depth before, so I had to look up how it works.
According to [GH Actions repo](https://github.com/actions/checkout/blob/f548e57e544e1ff5a4c46bf1e1b8685f8e4a348a/action.yml#L74), the value `0` is a magic number to have the full commit history, while numbers n>0 are used for the last n commits. Having the last 0 commits would otherwise be a very bad idea ^^

So my points are:
- is my understanding here correct?
- should we add a brief explainer like "get full commit history"?

.github/workflows/translator-alert.yml
```diff
@@ -0,0 +1,268 @@
...
+            
+            if (!category) {
+              core.setFailed("Discussion category not found.");
+              return;
+            }
+            
+            core.setOutput("repository_id", data.repository.id);
+            core.setOutput("category_id", category.id);
+
+      - name: Determine release title
```
moritz-gross: this step here makes it so that you have one discussion title per release as I understand it?
And later, in the step "Find or create release discussion", you either create or re-use that discussion.

We could add a brief explainer at the start of the file that we group the alerts in the discussions by release.

NSoiffer:
@moritz-gross: you're right, I thought it would need an external JS file and didn't look at the bottom to see it is embedded in the action. I asked AI if it could write something in shorter as a shell script or python. It did come up with something shorter for part of it using a shell script; it said python is not well suited to the task. Overall, it said JS/GraphQL is the best approach.

Still, the question remains: is this noise for the translators or useful?

moritz-gross:
> Still, the question remains: is this noise for the translators or useful?

in the last big meeting, the answer was a clear yes, as I understood it

MartheGjelstad:
Yes, in the meeting we had about the release process (August 17), we agreed on this procedure where the translators are being alerted on the first and 15th every month. I think we should try this workflow. We can adjust it later if we want to recieve less frequent notifications about changes.

moritz-gross:
@Johannes-spsm what do you think about my suggestions in the review? They all sound pretty quick to do, so we can get this merged soon. I don't want to poke in someone else's PR (at least not without getting their POV first)

Johannes-spsm:
@moritz-gross I agree the comment explaining the cron schedule is a good addition. I believe the GitHub API integration is worth using JS for. This PR is ready to be merged imo.

PR #764, opened 2026-09-11
Add Polish math braille code

michaldziwisz:
Adds Polish as an eleventh braille code, next to Nemeth, UEB, CMU, Vietnam,
Russian, Swedish, French, Finnish, LaTeX and ASCIIMath.

The code follows *Brajlowska notacja matematyczna, fizyczna, chemiczna* (2nd ed.,
Krakow-Laski-Lodz 2011), the notation recommended by the Polish Ministry of
Education for teaching blind pupils. It descends from the Epheser/Marburg
tradition, so some signs will look familiar next to the German code, but the
spacing rules and several key signs differ.

## How the cell values were obtained

The guide is a PDF that renders braille with a subsetted font, so the text layer
gives the font's own byte values rather than braille. Every cell in this PR was
therefore read from the GLYPHS: the dot pattern was recovered from each glyph's
bounding box and mapped to the Unicode braille block, and the mapping was then
cross-checked against two codes already in the repo (letters and digits agree with
Swedish, the number sign with UEB).

Each entry and rule carries the guide's page number in a comment, so anything here
can be traced back to its source.

## What is covered

- alphabet including the Polish diacritics, and digits
- 48 Greek letters, with the guide's single lower-case and upper-case Greek signs
  rather than a capital sign stacked on a Greek one (p. 7)
- operators, relations, negated relations, set theory, logic, geometry
- fractions (short and full form), roots including graded roots, powers and
  indices with lowered digits (p. 33)
- function names as contractions, trigonometry with the arcus form, limits,
  logarithms and the base written as an upper left index (pp. 36, 47, 52-56)
- absolute value, divisibility, brackets in four pairs, punctuation after numbers
- vectors, matrices and determinants, the projector technique
- units of measurement via the unit sign (pp. 69-72)
- non-structural chemistry: element symbols, molecules, ions (pp. 77-79)

62 tests in `tests/braille/Polish/polish.rs`, run as part of the braille suite.

## What is NOT covered, on purpose

- **Structural chemistry** (pp. 92-101, about 17% of the guide): ring compounds and
  bond diagrams are two-dimensional layouts, and MathCAT translates MathML
  linearly. Out of scope rather than half-done.
- **Full six-group spacing precedence** (pp. 3, 58-59). The guide classifies every
  sign twice - once for its left side (A/B/C) and once for its right (A'/B'/C') -
  and gives an order for conflicts. Implemented: the group-B blank on the left of
  relations and operators, and rule 1 for group C', which holds structurally
  because the blank is only ever emitted on the left. The general pairwise
  resolution is not implemented; the measured group membership for the signs we
  could establish is recorded in `definitions.yaml` so it need not be redone.
- **Roman numerals.** Page 6 writes them with a single capital sign for the whole
  run, but p. 57 states the general rule - the sign is repeated for every letter -
  and the guide's own PLN, geometry points and vectors follow it. Handling this
  needs Roman numerals to be RECOGNISED as such; collapsing every run of capitals
  breaks ordinary uses, so the limitation is documented in the test file instead.
- `get_braille_chars` reuses the UEB implementation, as Swedish does. The Polish
  letter and number key signs are applied in `polish_cleanup`.

## Notes for review

`src/braille.rs` gains only the registration (the `BrailleCode` impl, the cleanup
function and the indicator table) - the diff against main has no deleted lines
anywhere in the tree.

`polish_cleanup` collapses the capital-plus-letter and Greek indicator pairs before
the indicator table runs, because the Polish code has ONE sign for an upper-case
letter rather than a capital sign added to a lower-case one.

Spacing was the hardest part to get right, and the first version was wrong: blanks
were emitted symmetrically on both sides of every operator. Reading the guide's
glyphs over the whole document shows the asymmetry - for the equals sign, a blank
on the left in 79% of 380 occurrences and on the right in 3%; for plus, 88% and 9%;
for minus, 74% and 9%. Counting what follows a relation explains why: usually a
number sign or a letter sign, which is exactly the group C behaviour of rule 1.

### DIFF
--- a/Rules/Braille/Polish/Polish_Rules.yaml
+++ b/Rules/Braille/Polish/Polish_Rules.yaml
@@ -56,6 +56,130 @@
   match: "not(*)"
   replace: [t: "W"]
 
+# Absolute value / modulus (p. 12). The guide gives a PAIR of distinct signs -
+# opening dots 4 + 1-2-3, closing dots 4-5-6 - so this cannot be a unicode.yaml
+# entry for "|" (one character, two different meanings by position). Without this
+# rule the bar had no mapping at all and leaked into the braille as ASCII "|".
+# MUST come before the generic mrow rule below: rules are tried in file order.
+-
+  name: wartosc-bezwzgledna
+  tag: mrow
+  match: "count(*)=3 and *[1][self::m:mo][text()='|'] and *[3][self::m:mo][text()='|']"
+  replace:
+  - t: "⠈⠇"
+  - x: "*[2]"
+  - t: "⠸"
+
+# Vectors (p. 49). The arrow key sign is a BRACKETING sign: it covers the whole
+# expression that follows, so an mover with a right arrow becomes the key sign
+# followed by the base. Verified against the guide's own examples, e.g.
+# "wektor AB = -wektor DC" -> dots 4-5 + 2-5 + 2 then the letters.
+-
+  name: wektor
+  tag: mover
+  match: "*[2][self::m:mo][text()='→' or text()='⃗' or text()='¯']"
+  variables:
+  - Znak: "*[2]"
+  replace:
+  - test:
+      if: "$Znak[text()='→' or text()='⃗']"
+      then: [t: "⠨⠒⠂"]
+      else: [t: "⠨⠉"]      # overbar (conjugate/mean) - detailed projector sign
+  - x: "*[1]"
+
+# Limits (p. 56). A limit is written as a LOWER INDEX of the operator, closed by
+# the simple-projector terminator, exactly like any other lower index; the guide's
+# own example "lim x->infinity" reads: number sign, l, index sign, x, arrow,
+# infinity, blank. So munder over a limit-like operator reuses the index rule.
+-
+  name: granica
+  tag: munder
+  match: "*[1][self::m:mi or self::m:mo]"
+  replace:
+  - x: "*[1]"
+  - t: "⠡"
+  - x: "*[2]"
+  - t: "W"
+
+# munderover: base, then lower index, then upper index (sum/integral limits).
+-
+  name: granice-dolna-gorna
+  tag: munderover
+  match: "."
+  replace:
+  - x: "*[1]"
+  - t: "⠡"
+  - x: "*[2]"
+  - t: "W"
+  - t: "⠬"
+  - x: "*[3]"
+  - t: "W"
+
+# Matrices and determinants (p. 46). The guide brackets a matrix with dots 1-2-6
+# ... 3-4-6 and a determinant with dots 1-2-3 ... 4-5-6; rows are separated by a
+# blank cell and so are the entries within a row. Read off the guide's own worked
+# example, a 2x2 matrix and its determinant.
+-
+  name: macierz
+  tag: mtable
+  match: "."
+  replace:
+  - t: "⠣"
+  - x: "*"
+  - t: "⠜"
+
+-
+  name: wiersz-macierzy
+  tag: mtr
+  match: "."
+  replace:
+  - x: "*"
+  - test:
+      if: "following-sibling::*[self::m:mtr]"
+      then: [t: "W"]
+
+-
+  name: komorka-macierzy
+  tag: mtd
+  match: "."
+  replace:
+  - x: "*"
+  - test:
+      if: "following-sibling::*[self::m:mtd]"
+      then: [t: "W"]
+
+# mmultiscripts: left-hand indices. The guide devotes a section to them (p. 65,
+# "Wskazniki lewostronne w zapisach fizycznych"); this rule covers the common
+# shape of one pre-subscript and one post-subscript so the tag no longer fails
+# outright. The full left-index grammar is NOT implemented (see definitions.yaml).
+-
+  name: wskazniki-obustronne
+  tag: mmultiscripts
+  match: "."
+  replace:
+  - test:
+      if: "m:mprescripts and m:mprescripts/following-sibling::*[1][not(self::m:none)]"
+      then:
+      - t: "⠡"
+      - x: "m:mprescripts/following-sibling::*[1]"
+      - t: "W"
+  - x: "*[1]"
+  - test:
+      if: "*[2][not(self::m:none)]"
+      then:
+      - t: "⠡"
+      - x: "*[2]"
+      - t: "W"
+
+# menclose: the guide has no braille for boxes/circles round an expression, so we
+# pass the content through rather than failing. Reported as a gap, not a feature.
+-
+  name: menclose-przezroczysty
+  tag: menclose
+  match: "."
+  replace:
+  - x: "*"
+
 -
   name: default
   tag: mrow
@@ -132,7 +256,8 @@
   match: "text()='sin' or text()='cos' or text()='tg' or text()='ctg' or
           text()='sec' or text()='cosec' or text()='arcsin' or text()='arccos' or
           text()='arctg' or text()='arcctg' or text()='arcsec' or text()='arccosec' or
-          text()='min' or text()='max' or text()='const' or text()='sgn'"
+          text()='min' or text()='max' or text()='const' or text()='sgn' or
+          text()='lim'"
   replace:
   - test:
     # trigonometric, p. 52: prefix + one cell for the function
@@ -168,6 +293,9 @@
       then: [t: "⠼⠭"]
     - else_if: "text()='const'"
       then: [t: "⠼⠅"]
+    # limit, p. 56: the guide writes "lim" as number sign + l
+    - else_if: "text()='lim'"
+      then: [t: "⠼⠇"]
     - else:
       - t: "⠼⠎"
 
--- a/Rules/Braille/Polish/Polish_Rules.yaml
+++ b/Rules/Braille/Polish/Polish_Rules.yaml
@@ -1,0 +1,391 @@
+---
+# Polish math braille rules
+#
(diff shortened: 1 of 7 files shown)


NSoiffer (commented):
@michaldziwisz: I have made substantial progress on Polish braille, but only have something like 85% of the tests correct. See the "polish" branch. Take the tests from tests/braille/Polish/spec.rs in that branch and see how many of the 458 tests pass.

michaldziwisz:
Thanks for pointing me to `daisy/polish` - that changed what this PR should be.

## The number you asked for

**73 of the 458 active tests in your `tests/braille/Polish/spec.rs` pass against
the rules in this PR.** 385 fail.

Validity control, because the bare number would be worthless: I ran the same file
on your branch in a separate worktree and got **394/458**, i.e. exactly the 64
failures your commit message mentions. So the tests are runnable and
discriminating, and 73 is our real state, not a config artefact.

Two things I had to add to measure at all (not part of the rules, scaffolding only):
a `Polish_BrailleLevel` entry in `Rules/prefs.yaml` and `"Polish" => "pl"` plus
`set_preference("Polish_BrailleLevel", ...)` in `tests/common/mod.rs`, copied from
your harness. Without them 115 failures were "unknown preference", not braille.

## What I concluded from that, and what I propose

Your implementation dominates ours. Our 73 passing tests are a strict subset of
your 394 - **there is not one test that passes here and fails on your branch**
(checked with `comm` after every fix). Where we differ, I read the source guide
and you were right in 326 of 410 cases; I did not find a single case where your
expectation contradicts the guide.

So I am not proposing to bring our rules up to your spec. Instead I went through
your 64 failures and audited your tables against an independent reading of the
guide. That is the one thing we have that you don't: the braille in
"Brajlowska notacja matematyczna, fizyczna, chemiczna" is a **font, not Unicode**,
so I decoded the glyph outlines into a cell map (62 cells, agreement 0.99 on
891 occurrences of the most frequent one) and can quote what the document
actually prints, cell by cell.

**Audit result on your tables: clean.** No cell outside the guide in
`unicode.yaml`, `unicode-full.yaml`, `Polish_Rules.yaml` or `definitions.yaml`.
(My first run flagged `⠐` and `⣿`; both were my tool's fault - `⣿` is the end of
the range `⠀-⣿`, and `⠐` is an internal indicator. Mentioning it so you don't
have to re-check.)

## Six of your 64 failures cannot be fixed in code

`spec.rs` contains pairs of tests with **byte-identical input MathML and
contradicting expectations** - one passes, the other fails. Where I could check
it against the guide, it prints both forms:

| failing | passing twin | difference | guide |
|---|---|---|---|
| `fractions_p27_3b` | `fractions_p27_3a` | blank before the multiplication dot | p. 27: "Znak mnozenia - kropka, moze byc zastosowany z odstepem lub bez" - the same formula is printed twice, once each way |
| `fractions_p28_6a` | `fractions_p28_6b` | full vs short fraction form | p. 28 prints formula [6] both ways, captioned |
| `physics_p66_5a` | `physics_p66_5` | full vs short fraction form | same rule, p. 25 |
| `physics_p67_2a` | `physics_p67_2` | fraction-end sign + blank | same |
| `physics_p68_1` | `physics_p68_1a` | full vs short fraction form | same |
| `physics_p64_9` | `physics_p65_1` | stray line-break sign | see below; you wrote `// or` between them yourself |

**Your real ceiling is 452, not 458.** You already handle this correctly elsewhere
- `physics_p61_1/1a/1b` and `p61_2/2a/2b` have the variants under `#[ignore]`.
Applying that consistently takes 64 failures to 58 without touching a rule.

Related: in the fraction pairs your engine emits the **short** form, and the guide
calls it the shorter, space-saving one (p. 25, p. 27). Your default looks right
to me.

## Two expectations that are wrong, not the code

- **`limits_p56_3`**: a stray ASCII `X` at the end of the expected braille
  (`...⠶⠼⠁X`). Your output matches exactly once you drop it. I checked all 450
  expectations in the file: this is the **only** non-braille character anywhere,
  so it is an isolated typo, not a pattern.
- **`scripts_p31_14`** `(1/3)^-4`: the test wants `⠼⠙` (number sign + 4), your
  engine emits `⠲` (lowered 4). Guide p. 31 rule 1: integer exponents are written
  "cyframi obnizonymi bez znaku liczbowego". Its own examples decode as
  `b^4 → ⠠⠃⠬⠲`, `5^4 → ⠼⠑⠬⠲`, `a^-3 → ⠠⠁⠬⠤⠒`. The minus doesn't change it.
  **Your engine is right here.**
- **`physics_p64_9`** and **`reactions_p79_1`** both carry a line-break sign into
  the expected string. The guide breaks these examples across two braille lines;
  p. 65 says the 4th point goes at the end of the line **and repeats at the start
  of the next**. Decoding both lines of p. 64 gives your expectation plus one
  extra `⠈`; for p. 79 (photosynthesis) the two decoded lines equal the
  expectation character for character. The engine can't know the line width, so
  it can't produce these.

## Where the remaining failures actually come from

Grouped by cause, with the guide page. Happy to open issues or a PR for any of
these - tell me which you'd rather have.

1. **Right-hand script nesting opens a compound projector** (14 failures).
   `concentration_p87_4a` `c_{p1}`: decoding the same formula from p. 87 gives a
   bare `⠡`, no projector signs at all.
2. **No literary bracket** (12). p. 12 has a separate bracket `⠠⠶…⠠⠶` for
   non-mathematical asides; without it `5−3=2 (bo 2+3=5)` reads as "…=7 przez 5
   z 13". **We don't have this either** - shared gap, not a fault of yours.
3. **Missing blank before an operator** (31 - your largest group). The guide calls
   omitting it "powazny blad zapisu" (p. 10): `5 + x` without the blank reads as
   the fraction 5/6 followed by x. Your engine emits `⠈` (dots 4) there, which
   isn't a substitute for a blank cell.
4. **Fraction start/end signs emitted in short form** (`physics_p61_1a`,
   `physics_p61_2a`, `physics_p68_5`). p. 25 rule 1: they're omitted iff neither
   numerator nor denominator contains a blank. `physics_p61_1a` also loses the
   `⠨` before the second R. (`physics_p66_5a` looks like the same thing but is
   really one of the unresolvable pairs above.)
5. **Chemistry: short vs full molecule form** (4: `hydrates_p80_5a`,
   `reactions_p86_3`, `organic_cmpds_p88_2`, `structural_formulas_p91_1`).
   p. 77: the short form (molecule key `⠸` once) is allowed **only if every
   symbol in the molecule is single-letter**; otherwise the element key `⠨`
   before each. So in `Al₂O₃·H₂O` the two-letter `Al` forces the full form for
   the whole formula, including the `H₂O` part. Decoded from the guide:
   `KOH → ⠸⠅⠕⠓`, `HCOOH → ⠸⠓⠉⠕⠕⠓` (short) / `⠨⠓⠨⠉⠨⠕⠨⠕⠨⠓` (full).
   One decision at molecule level, not four bugs.
6. **Inverse function read as a power** (`functions_p40_12`, `scripts_p34_2`).
   `⠬` is the exponent key, `⠌` the upper-index key. p. 40 prints exactly the
   test's formula with `⠌`, because the `-1` on a function name isn't
   exponentiation. **`scripts_p34_2` is the one place we're worse than you**: we
   make the same mistake and add spurious brackets on top.
7. **Digit-group separator** (`physics_units_p72_2`). Decoding `1 kWh = 3 600 000 Ws`
   from the p. 72 table gives your expectation character for character:
   `⠼⠁⠻⠅⠨⠺⠓⠀⠶⠼⠉⠄⠋⠚⠚⠄⠚⠚⠚⠻⠨⠺⠎`. p. 4 forbids a blank as the separator.
8. **Electron mark** (`electron_p84_3a`). p. 84 allows "znaczek" of either a bar
   (dots 25) or a minus (dots 36) and prints both. Your `3b` (`msup`, minus)
   passes, `3a` (`mover`, bar) fails because the engine returns the same for both.

## One warning, from our own failed attempt

Seven of your 64 differ from your engine's output **only** in blanks and `⠈` -
the cell content is identical (`date_p9_4`, `fractions_p25_6`,
`fractions_p27_3b`, `ionic_bonds_p83_1`, `limits_p56_5`, `logarithms_p47_3`,
`prob_comb_p55_2`). Cheapest group to close, and all of it is item 3 above.

I tried exactly that on our own rules today and **failed, with a measurement**:
three approaches (a `U+00A0` entry in `unicode.yaml`, reading
`data-previous-space-width` the way Nemeth/UEB/Swedish do, and a branch in the
`mo` rule) gave +3 tests and −3 tests, net zero. The last one moved the blank to
the wrong place (`logarithms_p47_3`: a blank appears after `ln`, where it must
not). I reverted all of it.

The reason, I think, is that the guide's spacing isn't a per-sign property: p. 58
defines six groups A/B/C/A'/B'/C' with precedence C > B > A deciding whether a
blank must, must not or may appear on each side. Patching individual signs moves
the error rather than removing it. That chapter is unimplemented in our rules and
I suspect it's what's behind your 31.

## What I'd like to do with this PR

Your code is the better base - 1302 lines of tables plus 921 of Rust against our
569, and it wins on the merits where we disagree. Options, your call:

1. Close this PR, and I contribute to `daisy/polish` instead: fixes for the causes
   above, plus our 62 tests as supplementary coverage (they're written from the
   guide independently of your spec, so they exercise different examples).
2. Keep it open only as the carrier for the cell map / decoding script, if you
   consider that useful for checking tables against the printed guide.

Either way, happy to keep decoding specific formulas from the guide for you -
it's mechanical now, and it settles "what does the document actually print"
questions without argument.

PR #766, opened 2026-09-14
de, el: use $CommandOffset instead of string-length($Prefix) in navigate.yaml

### LINKED ISSUE #740: de and el navigate.yaml still use string-length($Prefix), which breaks as soon as their prefixes are translated
`ec36e057` replaced `string-length($Prefix)+1` with `$CommandOffset` in
`navigate.yaml`, because `$NavCommand` is always English (`ZoomIn`, `MoveNext`)
while `$Prefix` is the spoken, translated word. `de` and `el` still carry the old
formula:

```
lang   string-length($Prefix)+1   $CommandOffset   Prefix words
de                            9                0   zoom, move, read, describe
el                            9                0   zoom, move, read, describe
en fr hu nb sv pl             0                9   (translated)
```

Right now both work, but only by coincidence: their `Prefix` values are still the
untranslated English words, so `string-length("move")` happens to equal the length
of the `Move` stem. The first translator who localizes them — the obvious thing to
do with a user-facing string — silently breaks every navigation announcement in
that language.

This is not hypothetical; it is exactly what `pl` was suffering from before #738.
With `Prefix: 'przejdź'` (7 chars) against the `Move` stem (4):

```
ZoomIn        -> ""          (want "In")
ZoomOutAll    -> "ll"        (want "OutAll")
MoveNext      -> "t"         (want "Next")
DescribeNext  -> "ibeNext"   (want "Next")
```

All 16 branches were dead, so users heard "przejdź; do mianownika" with no
direction word at all. Nothing failed: there were no navigation tests for `pl`,
and the 600+ speech tests were green throughout.

Two things might be worth doing:

1. Port `$CommandOffset` to `de` and `el` now, while their prefixes are still
   English and the change is a no-op that can be verified as such.
2. Since the same trap can reappear in any new language, it might be worth
   comparing against `$NavCommand` directly, the way `ru` does
   (`$NavCommand = 'ZoomIn'`). More verbose, but it cannot be got wrong by
   choosing a different word length.

I am happy to send a PR for point 1 if you would like it — it is mechanical, and
I would keep it to those two files.

For reference, `es`, `fi`, `id`, `vi` and `zz` do not use this construct at all.

michaldziwisz:
Closes #740 point 1.

`ec36e057` replaced `string-length($Prefix)+1` with `$CommandOffset`, because
`$NavCommand` is always English (`ZoomIn`, `MoveNext`) while `$Prefix` is the
spoken, translated word. `de` and `el` were the last two files still carrying the
old formula; this ports the change to them and touches nothing else.

**This is deliberately a no-op today**, which is why it is a good moment to do it.
Both languages still use the untranslated `zoom`/`move`/`read`/`describe` as their
`Prefix`, so `string-length("move")` happens to equal the length of the `Move`
stem and all 18 branches match. The first translator who localizes those strings
loses every direction word, silently.

### What changed

- 9 uses of `substring($NavCommand, string-length($Prefix)+1)` → `$CommandOffset`
  in each file;
- the 4 `set_variables` calls now also set `CommandOffset` (5 for Zoom/Move/Read,
  9 for Describe), same values as `en`.

Translations are untouched: every line in the diff is one of those two shapes.
After the change the offsets and the list of 9 branches (`In`, `InAll`, `Out`,
`OutAll`, `Next`, `Previous`, `Current`, `LineStart`, `LineEnd`) are identical to
`en/navigate.yaml`.

### Verification

Checked on the engine rather than only on paper, because the interesting property
here is "nothing changes":

- `Languages::de` → **209 passed, 0 failed**, and **the same 209/209 with the
  change reverted**. So the existing German tests do not discriminate this at all,
  which matches the report: there are no navigation tests for `de` or `el`.
- To get a real measurement I wrote four throwaway navigation tests for `de` on
  the existing `init_nav` / `do_navigate_command` harness (the shape used by
  `en` and `nb`), with the expected strings **read out of the engine** rather than
  written by hand. Result: 4/4 with the change, 4/4 without it — the no-op is
  confirmed on live output.
- Validity control, so those tests are not vacuous: translating the prefixes
  (`zoom` → `zoomen`, `move` → `gehe`) makes the OLD formula produce
  `"zoomen; in zähler; 1"` — the direction word is simply gone — while the new one
  produces `"zoomen in; in zähler; 1"`. Same simulation over all 15 navigation
  commands: the old formula breaks 7/15 for German prefixes and 15/15 for Greek
  ones.

I kept those tests out of this PR to hold it to the two files I promised in the
issue. Happy to send them separately if you want `de` navigation pinned.

### One thing I noticed while measuring

**Greek has no tests running at all.** `tests/Languages/el/` exists with 11 files,
but `el` is not declared in `tests/languages.rs`, so cargo filters everything out
(`0 passed; 5139 filtered out`). Current spread: pl 625, hu 589, nb 583, en 580,
fr 518, ru 513, sv 503, fi 488, zh 290, de 209, pt 174, vi 19, id 16, **el 0**.
So for `el` this change rests on the structural check only. Say the word and I
will open a separate issue for that — it looked like an oversight rather than a
decision.

Point 2 of the issue (comparing `$NavCommand` directly, the way `ru` does) is not
part of this PR; it changes all languages, so it deserves its own decision.


NSoiffer:
This looks good.

I need to look into the `el` tests. You are correct that it should be part of languages.rs. However, the `el.rs` file has all the lines commented out. Hopefully they all pass. I'll give them a try.

PR #774, opened 2026-09-19
Refactor: Simplify match return logic and improve string interpolation readability

moritz-gross:
small refactor I found while skimming a bit through the codebase.


NSoiffer:
Seems a bit like rearranging chairs on the deck of ship. But I'll approve it.

PR #775, opened 2026-09-19
Refactor: Replace custom path utilities with `Path` API and update zip methods accordingly

moritz-gross:
- reduces LOC by 10
- having proper types instead of str is safer imo

Side note: This file also contains some unused imports. should I remove them as well?

and I'm surprised somewhat that the previous implementation is not buggy in some way?? hard-coding a path separator like "/" is usually a bad idea from my experience, given the kinds of this operating systems are doing.


NSoiffer:
@moritz-gross Except for some special cases, "/" works everywhere in Rust. There is `MAIN_SEPARATOR`, but is obviously bulkier and less readable.

Please test this with `feature = "include-zip"`. There is code in `shim-filesystem.rs ` that wan't to have "/"s due to dealing with zip files.

moritz-gross:
`cargo test --features include-zip` passes

(I'm getting this warning though, which is fine I guess:  _warning: methods rule_files_cache_is_empty, definitions_files_cache_is_empty, and definitions_files_cache_path are never used_ )

I've not used the flag `--features include-zip` before.

PR #776, opened 2026-09-19
Refactor navigation state management: unify position and command stacks into a single history vector

moritz-gross:
I have a hard time understanding what's happening in some files, so I'll keep suggesting some ways in how I think it can be done a bit smoother. Either my proposals are good enough to be merged, or I'll find out how I am misunderstanding the code ;)

regarding your comment in the code: I'm 99% sure Vec is the correct datastructure for this. Ideally, you want to use a contiguous memory layout, as opposed to pointer chasing in a linked list.
See for example also [VecDeqeue](https://doc.rust-lang.org/beta/std/collections/struct.VecDeque.html) for some other use cases where we might want to manipulate both ends. But for this, Vec seems perfect. Its usage as a stack is also documented in https://doc.rust-lang.org/std/vec/struct.Vec.html


moritz-gross:
the failing test doesn't reproduce locally, so something about the CI setup is off.
See relevant part of the CI logs:

> failures:
> 
> ---- Languages::en::alphabets::cap_cyrillic stdout ----
> Error: MathCAT crash! Please report the following information: 'MathCAT crash! Please report the following information: 'MathCAT crash! Please report the following information: 'RefCell already borrowed' at src/speech.rs:2811:0
> ' at tests/common/mod.rs:56:0' at tests/common/mod.rs:61:0
> 
> 
> failures:
>     Languages::en::alphabets::cap_cyrillic
>

PR #777, opened 2026-09-19
Fix audit-ignore rule scoping

### LINKED ISSUE #742: bug: `parsers.py` can misattribute the `#audit-ignore` comment
see PR https://github.com/daisy/MathCAT/pull/728

as recently spotted by @yasumorishima, the placement `# audit-ignore` can have some misleading behaviour, as the following example is not properly recognized:
```
# audit-ignore
- name: per-fraction
  tag: fraction
  match:
  - ...
```

but on the other hand, this works: 
```
- name: per-fraction # audit-ignore
  tag: fraction
  match:
```

In the design of the `audit-ignore` tool, I didn't give this part much attention, but now thinking about it, placing the comment directly in front of the affected rule looks the cleanest to me.

My proposal is to implement this and document it as a best practice in the docs.

thoughts and opinions on this?

Kostenkov-2021:
Adjust the audit-translations parser so `# audit-ignore` markers immediately above a rule apply to that rule, while still supporting existing inline markers inside a rule block. This also clarifies the documented behavior and adds regression tests for leading, inline, and first-item ignore cases.
• Closes #742


PythonScripts/audit_translations/parsers.py
```diff
@@ -149,13 +149,20 @@ def format_tag(tag_value: Any) -> str | None:
...
-        return blocks
+        return []
+
+    block_starts = starts.copy()
     for idx, start in enumerate(starts):
-        end = starts[idx + 1] if idx + 1 < len(starts) else len(lines)
-        blocks.append("\n".join(lines[start:end]))
-    return blocks
+        lower_bound = starts[idx - 1] + 1 if idx else 0
+        while start > lower_bound and (not lines[start - 1].strip() or lines[start - 1].lstrip().startswith("#")):
```
moritz-gross: I think this is very hard to read.
At first, I missed that the whole part after "and" is grouped by "( ... )".

PythonScripts/audit_translations/parsers.py
```diff
@@ -149,13 +149,20 @@ def format_tag(tag_value: Any) -> str | None:
...
+            start -= 1
+        block_starts[idx] = start
+
+    return [
+        "\n".join(lines[start : block_starts[idx + 1] if idx + 1 < len(block_starts) else len(lines)])
```
moritz-gross: also a bit too crowded to be read easily.  It's hard to tell which [ ] belong together.
Extracting `end_line` as a variable, or similar, would help here

moritz-gross:
logic looks correct, tests make sense and cover the important parts.
if we make the changes easier to follow, then we're good to go in my opinion

moritz-gross:
```
>         content = """- name: first
>   tag: mo
>   match: "."
> 
> # audit-ignore: the following placeholder intentionally differs
> # from the source-language rule.
> - name: second
>   tag: mi
>   match: "false()"
> """
```

on a side-note, I'm not really happy with how the formatting of the first vs. second line of these YAML strings looks.
As I've used the same ugly style myself, it would be unfair to criticize that here.
When I worked on it, I didn't want to immediately do a full line break after """ at the start, but I think this would be visually the best option.

Or is there anything else I'm missing here?
Again, only semi-related to the PR, this only brought it up to my attention again.

yasumorishima:
Numbers for the trailing-comment case, measured on `main` (`49de6d67`) over `Rules/Languages`.

There are 1,140 comment runs between consecutive rules. Splitting them by whether a blank line sits above the run:

| shape | count | audit-ignore markers |
|---|---|---|
| blank line above the run (leading comment for the next rule) | 995 | 2 (both `ru`) |
| directly under the item's own `- ` line | 39 | 0 |
| touching the previous rule's last content line | 106 | 0 |

The middle group is worth calling out: `data.lc.item(idx)` reports the `name:` line, so for

```yaml
- # intervals are controlled by a ClearSpeak Preference ...
  name: ClearSpeak-intervals
```

`main` attributes those comment lines to the *previous* rule even though they are textually inside this item. This PR fixes 39 of those as a side effect.

The third group is the one flagged in #742. A predicate that keeps it with the previous rule, while still fixing the other two, is three branches when walking up from `starts[idx]`: stop and take the run if a blank line is above it; also absorb the item's own `- ` line (a line matching `^\s*-\s*(#.*)?$`) and stop; otherwise hand the run back to the previous block.

I ran that over the tree and it yields **exactly the same set of ignored rules as this PR** — 0 differences — both differing from `main` only by the two `ru` placeholders (296 in rule files, plus the 2 in `hu/unicode.yaml`, which is parsed by the unicode path, = 298).

One more shape, whichever way this goes: a marker written on the dash line itself (`- # audit-ignore`) still lands on the previous rule, because the upward walk stops at the dash. 146 dash lines carry a comment today and none of them carry a marker, so this is a README line rather than a bug.

moritz-gross (approved):

moritz-gross:
thanks for your work. merged!

yasumorishima:
Thanks for merging. I measured the merged `main` (`374ea588`) against the merge base (`5c8117e9`) afterwards, since I had the numbers half-collected; posting them here in case they are useful as a record. Nothing below asks for a change.

The corpus is the 174 non-`definitions.yaml` files under `Rules/Languages`, 172 of which parse into rules (`el/unicode.yaml` still raises `DuplicateKeyError`, one file has no rules). The `Rules` tree is byte-identical on both sides, so the comparison is like for like.

**The attribution moves by exactly one pair, and it moves where it should.** Rules whose block satisfies `has_audit_ignore`, counted through `parse_yaml_file`: 298 before, 298 after; the corpus holds 298 `# audit-ignore` lines. The two sets differ only in `Rules/Languages/ru/navigate.yaml`:

| marker | suppressed before | suppressed now |
|---|---|---|
| lines 1369-1370 | `move-next-none`, `tag: none` (line 1348) | `move-next-none`, `tag: [none, mprescripts]` (line 1371) |
| lines 1398-1399 | `move-previous-none`, `tag: none` (line 1377) | `move-previous-none`, `tag: [none, mprescripts]` (line 1400) |

The marker itself reads "parity placeholder for English tag list without changing the existing Russian none-specific rule above", so the rule below it is the one it was written for. End to end, `audit-translations ru --file navigate.yaml` goes from 214 to 206 rule issues (rule differences 34 to 26): the eight issues raised against the two placeholders disappear, and diffing the two reports shows removals only, nothing newly reported.

The dash-line form is fixed as well. With the marker on the item's own dash line

```yaml
- # audit-ignore
  name: second
```

the merge base attributes it to the preceding rule and main now attributes it to `second`. A comment run separated by a blank line lands on the following rule, and a run touching the preceding rule's body stays with that rule, as intended.

One boundary worth writing down, with no action needed today: a marker on an indented line *under* the dash line, with no blank above it, still goes to the preceding rule.

```yaml
-
  # audit-ignore
  name: second
```

(the same if the dash line already carries a comment). In `Rules/Languages` today, 144 rules have their `name:` directly under a dash line and 37 have indented comments between the dash line and `name:`; neither group contains an `audit-ignore` marker, which is why the 298 counts above are unchanged. The README documents the blank-line form; the dash-line form now works too, if it is ever worth a sentence there.

PR #782, opened 2026-09-20
use elements.rs as chemistry database

moritz-gross:
while wandering some more through the codebase, I saw that we have a whole periodic table laying around here.
I think this is something we'd better hand off. 
I found the neat crate [elements_rs](https://crates.io/crates/elements_rs) for this. In particular, having categories like `ElementCategory::NobleGas` (not to speak of `AlkalineEarthMetal` etc.)

next to refactoring code, I this also uncovered a buggy test re Ammonia: The electronegativity of Nitrogen is 3.04, and that of Hydrogen is 2.20, so NH_3 is in descending order.

most likely, there is way more to hand off in this file, but this looks like an easy-to-do thing for now.


moritz-gross:
TODO: check difference in MathCAT size, and how to judge such a thing

moritz-gross:
binary size seems to actually shrink ?!?.

Running this: 
```
git switch main
cargo clean
cargo build --release --lib --locked
cp target/release/liblibmathcat.so /tmp/mathcat-main.so

git switch use-elements-rs-as-chemistry-database
cargo clean
cargo build --release --lib --locked
cp target/release/liblibmathcat.so /tmp/mathcat-elements.so

stat -c '%n: %s bytes' \
    /tmp/mathcat-main.so \
    /tmp/mathcat-elements.so
```

gives the result 
```
/tmp/mathcat-main.so: 605344 bytes
/tmp/mathcat-elements.so: 604152 bytes
```

The "eliminating-dead-code" step you mentioned is *Link-Time Optimization* as I just learned?! Seems to work well then.

@NSoiffer have you ever used the [size](https://man7.org/linux/man-pages/man1/size.1.html) command for inspecting section sizes of binaries? That could be interesting to inspect it further, but I think this would be more of a fun side-project than actually pragmatic here.

So in summary, this PR seems good to go (I'll not merge it myself as I want your feedback if my way of checking the size seems reasonable)

Issue #739, opened 2026-08-31
Arity templates in IntentMappings emit the English word "of" in every language


michaldziwisz:
`intent_function_glue_before` in `src/infer_intent.rs` returns a hardcoded
English `"of"` for the last argument of an arity template:

https://github.com/daisy/MathCAT/blob/main/src/infer_intent.rs#L316-L330

```rust
// words has arg_count-1 entries for args 1..arg_count-1; last arg is preceded by "of"
if arg_index == arg_count {
    return "of".to_string();
}
```

So any non-English language that adopts the arity-template form gets an English
word in the middle of its output. Measured in Polish with
`"sum": "function=suma | po | od,do"`:

```
spoken:   suma po i of x
expected: suma po i, x        (or whatever the language's "of" equivalent is)
```

The binary-separator form (a single `|` option, e.g. `| divided by`) is not
affected, because the separator word itself comes from `definitions.yaml`.

Consequence today: a translator who follows the documented syntax gets a broken
result, and the working alternative is to *not* use arity templates. That is what
`hu` does — `"sum": "function=összeg"` with no template — and what I ended up
doing for `pl` in #738 rather than shipping mixed-language speech.

A fix would presumably read the word from the language's rule files, the same way
`IntentFunctionArgSeparator` resolves `comma`. I have not attempted it, since it
touches the engine rather than a translation, and I would rather not guess at the
intended design here.

Happy to test a fix against the Polish rules if that is useful.

NSoiffer:
I can't believe that I made that mistake :-(

I have a fix that involves defining the word to use in definitions.yaml. For technical reasons, reusing U+2061 (invisible function apply) in `unicode.yaml` won't work without some ugly hacks.

I'm still working through some checks. Once I check this in, I'd appreciate you testing with Polish.

NSoiffer:
@michaldziwisz : go ahead and try this for Polish. The line to change is
```
- FunctionApplicationWord: ["of"]
```

michaldziwisz:
Tested on Polish against `main` at a74a6e2d (crate 0.7.6-rc.4). It works — thank you.

**Setup.** Two changes in `Rules/Languages/pl/definitions.yaml`: added `- FunctionApplicationWord: ["z"]`, and restored the arity templates I had removed in #738, i.e. `"sum": "function=suma | przez | od,do"` and the same for `product`. No changes to `src/`.

**Measured speech** (identical in ClearSpeak and SimpleSpeak):

```
[...]

PR #783, opened 2026-09-21
pl: restore sum and product arity templates after the connector fix

michaldziwisz:
## Summary

Follow-up to #739 and the [Polish verification report](https://github.com/daisy/MathCAT/issues/739#issuecomment-5677846485).

Now that the engine reads the connector from the language definitions instead of hard-coding English "of", restore the Polish arity templates for `sum` and `product`:

- Define `FunctionApplicationWord: ["z"]`.
- Use `function=suma | przez | od,do` and `function=iloczyn | przez | od,do`.
- Update the four existing expectations that recorded the previous workaround.

For example, `sum($i,$x)` becomes "suma przez i z x" instead of "suma z i przecinek, x", and `sum($i,$n,$x)` becomes "suma od i do n z x". The corresponding product forms use "iloczyn". The one-argument forms remain unchanged.

The Polish wording follows the existing large-operator rules. This changes only `Rules/Languages/pl/definitions.yaml` and `tests/Languages/pl/definitions.rs`. No engine or braille changes; independent of #764.

## Verification

Rechecked against current `main` at `374ea588cd77edc1745b8a632da1d23ec6e9c71f` (crate 0.7.6-rc.4):

- Unmodified upstream: `cargo test --test languages Languages::pl -j2`: **625 passed, 0 failed**.
- With this change, the same suite: **625 passed, 0 failed**.
- With `--features no-unsafe`: **625 passed, 0 failed**.
- Negative control: reverted only the Polish definitions, retaining the updated expectations. `sequence_and_series_intents` failed on the intended difference: expected "suma przez i z x", actual "suma z i przecinek, x". Restoring the definitions makes it pass again.
- `git diff --check` passes. The diff contains only the two files listed above.


Rules/Languages/pl/definitions.yaml
```diff
@@ -298,6 +296,11 @@
     "chemistry-concentration": "function=; stężenie: stężenie: stężenie; koniec stężenia",
   }
 
+  # The connector word to say for functional 'IntentMappings' (in English, "of")
```
moritz-gross: a general question: how do we want to handle the language of comments in rule files?
This one is in English, but the change above is in Polish (I assume).

I'm not sure what makes more sense, or if should try to have a policy on this.
What do you think? Maybe we should coordinate with the other translators as well.

NSoiffer:
Thanks. I never would have guessed that "z" => "of".

PR #789, opened 2026-09-22
replace lamda with lambda

moritz-gross:
I only just learned that there are two forms of λ: lambda *and lamda*. At least in German, it should be with "b", and the same goes for Indonesian from what I could dig up.
It also exists as "lamda" in the Unicode names in `PythonScripts/BrailleMathCodes Repository.csv`, but this is correct from what I learned ?!

and a fun fact: [LamDA](https://en.wikipedia.org/wiki/LaMDA) is the OG Google LLM family, pre-Gemini. Was the first thing I thought of....


NSoiffer:
It looks like there is duplication in the unicode-full and unicode files. But that's for another day (feel free to get AI to find the duplication and remove it -- typically it should go away in unicode-full.yaml. There is another issue to swap around what goes where. I'm not sure if the program that does that handles duplication well.

AI says that it is written as "lambda" but it sounds more like "lam-da" (as in English). Hopefully the speech engine does this correctly.

